PR #39, opened 2018-10-25
Update eslint dependencies

agustito37:
Eslint is throwing some deprecated warnings even with the ecmaFeatures properly configured:

```
(node:31829) DeprecationWarning: [eslint] The 'ecmaFeatures' config file property is deprecated, and has no effect. (found in /project_folder/node_modules/eslint-config-airbnb/rules/react-a11y.js)
(node:31829) DeprecationWarning: [eslint] The 'ecmaFeatures' config file property is deprecated, and has no effect. (found in /project_folder/node_modules/eslint-config-airbnb/rules/react.js)
```

So I updated eslint dependencies and the problem was solved. Additionally I disabled some new rules that could be a little annoying:

[react/desctructuring-assignment](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md)
[react/jsx-one-expression-per-line](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-one-expression-per-line.md)
[prefer-promise-reject-errors](https://eslint.org/docs/rules/prefer-promise-reject-errors)
[operator-linebreak](https://eslint.org/docs/rules/operator-linebreak)
[object-curly-newline](https://eslint.org/docs/rules/object-curly-newline)
[implicit-arrow-linebreak](https://eslint.org/docs/rules/implicit-arrow-linebreak)


.eslintrc
```diff
@@ -62,11 +62,19 @@
     "keyword-spacing": 1,
     "indent": [1, 2, { "SwitchCase": 1 }],
     "max-len": 0,
+    "operator-linebreak": 0,
+    "object-curly-newline": 0,
+    "implicit-arrow-linebreak": 0,
+    "prefer-destructuring": 0,
```
bernabe9: I wouldn't disable this one. We always prefer to use destructuring.

For example, if I see the following code:
```javascript
const foo = object.foo
```

I will comment to use object destructuring.
```javascript
const { foo } = object
```

What downsides do you see to this rule?

agustito37: Sure, I have a rule, maybe is not an universal standard, but I only destructure when I use more than one field of that object. 

E.g. I have a `prop` `user` with `id` and `name` and I will use just the name:

I don't declare `const { name } = this.props.user`, I just use  `user.name`.

pedrozs: Right, that pattern is considered a warning for the rule you mention.
Just like you, I prefer to avoid the additional line that requires.

PR #48, opened 2019-01-16
Add loginPage tests

pedrozs:
Add test file for loginScreen and required dependencies.


tests/LoginScreen.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      user: {
+        username: null,
+        password: null,
+        id: 1,
+        email: userEmail,
+        firstName: null,
+        lastName: null,
+        clientPersonCount: 0,
+        countryCode: null,
+        clientPersons: []
```
elizabethlofredo: I think you can remove some values like: `clientPersonCount` and `countryCode` just leave `username`, `password` `email` or the ones you need and are generic. The rest depends on the project.

tests/LoginScreen.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      }
+    };
+
+    sessionService.saveUser = jest.fn(() => Promise.resolve());
+    emailInput = wrapper.find(('TextInput')).at(0);
```
elizabethlofredo: unnecessary double `()`

tests/LoginScreen.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    passwordInput = wrapper.find(('TextInput')).at(2);
```
elizabethlofredo: Same here

__mocks__/react-native-config.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  API_URL: 'https://mock.com/api',
+  ASSETS_URL: 'https://mock.com'
```
elizabethlofredo: necessary in the base?

pedrozs:
@elizabethlofredo Fixed your comments, thanks!

tests/LoginScreen.spec.js
pedrozs: Native is rendering 2 TextInputs per Field. Thats why i need to get the third one.

PR #49, opened 2019-01-18
Add signUp screen

pedrozs:
## Description
Adds a signup screen and corresponding test

## Screenshots
screen shot 2019-01-18 at 11 57 43 am

-----

screen shot 2019-01-18 at 11 57 58 am

### DIFF
--- a/src/components/user/SignUpForm/styles.js
+++ b/src/components/user/SignUpForm/styles.js
@@ -1,0 +1,9 @@
+import { StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  signUp: {
+    width: 230
+  }
+});
+
+export default styles;
--- a/src/containers/LoginScreen/index.js
+++ b/src/containers/LoginScreen/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Text, View } from 'react-native';
+import { Text, View, Button } from 'react-native';
 import { connect } from 'react-redux';
 
 import LoginForm from 'components/user/LoginForm';
@@ -8,19 +8,25 @@ import { login } from 'actions/userActions';
 import translate from 'utils/i18n';
 import styles from './styles';
 
-const LoginScreen = ({ login }) => (
+const LoginScreen = ({ login, navigator }) => (
   <View style={styles.container}>
     <Text style={styles.welcome}>
       {translate('SIGN_IN.title')}
     </Text>
     <LoginForm onSubmit={user => login(user.toJS())} />
+    <Button
+      title={translate('SIGN_UP.title')} onPress={() => navigator.push({
+        screen: 'reactnativebase.SignUpScreen'
+      })}
+    />
   </View>
 );
 
-const { func } = PropTypes;
+const { func, object } = PropTypes;
 
 LoginScreen.propTypes = {
-  login: func.isRequired
+  login: func.isRequired,
+  navigator: object.isRequired
 };
 
 LoginScreen.navigationOptions = {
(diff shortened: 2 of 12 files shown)


tests/SignUpScreen.spec.js
```diff
@@ -0,0 +1,175 @@
...
+      user: {
+        username: null,
+        password: null,
+        id: 1,
+        email: userEmail,
+      }
+    };
+
+    sessionService.saveUser = jest.fn(() => Promise.resolve());
+    emailInput = wrapper.find('TextInput').at(0);
```
matiasmansilla1989: is there any way to avoid finding by position `at(0)` and using the label name or something else?

pedrozs: A selector can be made from props. But many components have will have the same prop as it it's passed down many times. I'll see if I can do better.

pedrozs: Maybe we could use refs for this? Enzyme can return React components from Refs.

felipegarcia92: Actually, one of the first tests is confirming this input at position number 0 has the name `email`. 

If we fetch it using that name, the test wouldn't make sense. However, we could change it and make the test assert that it exists

fsarachu: I would say that using `at(0)` is just adding one more assert to the test, expecting the email input to be the first one. If that's valid or not depends on what we want to test.
However, I think we could later try to use some pattern like [Page Object](https://medium.com/@boriscoder/page-object-for-testing-react-js-components-49cddb59adb7) to solve this kind of issues and write more semantic tests.

matiasmansilla1989: @felipegarcia92 has a good point. 
I would try to avoid having tests failing just because the position of the inputs is changed. However, if the alternative solution is complicated or not the best one (I understand `refs` is this case), we can keep the current one. At the end, as Felipe mentioned, we have specific tests to check that.

src/components/user/SignUpForm/styles.js
```diff
@@ -0,0 +1,9 @@
+import { StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  signUp: {
+    width: 230
```
kevquincke: Mmm why do you need this fixed width? A horizontal padding would be better

src/containers/LoginScreen/index.js
```diff
@@ -1,26 +1,32 @@
...
   </View>
 );
 
-const { func } = PropTypes;
+const { func, object } = PropTypes;
 
 LoginScreen.propTypes = {
-  login: func.isRequired
+  login: func.isRequired,
+  navigator: object
```
kevquincke: This one should be required

src/components/user/SignUpForm/styles.js
pedrozs: I just copied the style from the login form. Should I change both?

kevquincke: Yeah I think. We shouldn't use fixed width unless necessary, you can achieve the same and better (responsive) results with an horizontal padding here

pedrozs: Ok, I will, but in another PR fixing both styles. Sounds ok?

kevquincke: Sounds good!

src/containers/SignUpScreen/styles.js
```diff
@@ -0,0 +1,17 @@
+import { StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'center',
+    backgroundColor: '#F5FCFF'
```
missx: We don't have a constants file to store this?

src/components/user/SignUpForm/styles.js
felipegarcia92: For me, it's okay if you do it on this PR.

PR #56, opened 2019-04-15
Replace React Native I18n with React Native Localize

### LINKED ISSUE #55: react-native-i18n dependency is deprecated
https://github.com/AlexanderZaytsev/react-native-i18n#%EF%B8%8F-this-package-is-deprecated

We should migrate to [React Native Localize](https://github.com/react-native-community/react-native-localize) which is the recommended alternative given by the author

mariosouto:
Resolves #55

### DIFF
--- a/__mocks__/react-native-localize.js
+++ b/__mocks__/react-native-localize.js
@@ -1,0 +1,36 @@
+const getLocales = () => [
+  { countryCode: 'US', languageTag: 'en-US', languageCode: 'en', isRTL: false },
+  { countryCode: 'EC', languageTag: 'es-EC', languageCode: 'es', isRTL: false },
+];
+
+const findBestAvailableLanguage = () => ({ languageTag: 'es', isRTL: 'false' });
+const getNumberFormatSettings = () => ({
+  decimalSeparator: '.',
+  groupingSeparator: ',',
+});
+
+const getCalendar = () => 'gregorian';
+const getCountry = () => 'ES';
+const getCurrencies = () => ['USD'];
+const getTemperatureUnit = () => 'celsius';
+const getTimeZone = () => 'Ecuador/Guayaquil';
+const uses24HourClock = () => true;
+const usesMetricSystem = () => true;
+
+const addEventListener = jest.fn();
+const removeEventListener = jest.fn();
+
+export {
+  findBestAvailableLanguage,
+  getLocales,
+  getNumberFormatSettings,
+  getCalendar,
+  getCountry,
+  getCurrencies,
+  getTemperatureUnit,
+  getTimeZone,
+  uses24HourClock,
+  usesMetricSystem,
+  addEventListener,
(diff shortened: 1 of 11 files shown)


mariosouto:
This tackles https://github.com/rootstrap/react-native-base/issues/55

android/settings.gradle
```diff
@@ -1,10 +1,10 @@
 rootProject.name = 'ReactNativeBase'
-include ':react-native-i18n'
-project(':react-native-i18n').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-i18n/android')
-include ':app'
```
mato2593: Did you test if the app runs in android without this?

mariosouto: @mato2593 Actually i removed that library so we don't need to specify that in gradle anymore. Is that what you meant?

mato2593: No, I meant the `include ':app'`, which is the main module of the app

mariosouto: @mato2593 Oh, i move that to the end of the file... i think it's the same thing right?

tests/setup-tests.js
```diff
@@ -2,4 +2,8 @@ import enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import 'isomorphic-fetch';
 
+jest.mock('react-native-localize', () => ({
```
fsarachu: How about moving this mock to the `__mocks__` directory as we do for `react-native-config`?

Example:
https://github.com/rootstrap/react-native-base/blob/master/__mocks__/react-native-config.js

Docs:
https://jestjs.io/docs/en/manual-mocks#mocking-node-modules

mariosouto: Yeah, seems better. I'll change it

android/settings.gradle
mato2593: I didn't see it 😅

tests/setup-tests.js
mariosouto: @fsarachu Done, thanks for the feedback!

__mocks__/react-native-localize.js
```diff
@@ -0,0 +1,36 @@
...
+const getCurrencies = () => ['USD'];
+const getTemperatureUnit = () => 'celsius';
+const getTimeZone = () => 'Ecuador/Guayaquil';
+const uses24HourClock = () => true;
+const usesMetricSystem = () => true;
+
+const addEventListener = jest.fn();
+const removeEventListener = jest.fn();
+
+export {
```
fsarachu: For consistency purposes, it would be better to add the inline `export` statement on the function definitions, like this:
```js
export const findBestAvailableLanguage = () => ({ languageTag: 'es', isRTL: 'false' });
```

Other than that, this looks great!

mariosouto: Done 👍

Issue #18, opened 2018-04-24
Different environment variables for development, staging and production


bernabe9:
This base allows to have one `.env` file to store the env variables, but it would be better if we can manage multiples env files for each environment.

Reference: https://github.com/luggit/react-native-config#different-environments

PR #57, opened 2019-04-24
Add environment file for staging and production environments

mariosouto:
Resolves https://github.com/rootstrap/react-native-base/issues/18

### DIFF
--- a/.env.defaults
+++ b/.env.defaults
@@ -1,0 +1,1 @@
+API_URL=https://api-example.com
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,8 @@ buck-out/
 
 # Constants location
 .env
+.env.staging
+.env.prod
 
 # Coverage
 coverage/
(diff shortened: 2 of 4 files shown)


.nvmrc
```diff
@@ -0,0 +1 @@
+v8
```
bernabe9: change it to `10.14.2`. It's the same that we are using in the react-redux-base(https://github.com/rootstrap/react-redux-base/blob/master/.nvmrc)

mariosouto:
@bernabe9 Yes, you can specify which `.env` you want to use from the command line or you add an schema to load the correspondent file from XCode. Maybe i'm missing some documentation on how to run the command

.nvmrc
mariosouto: Actually if we run this on Node 10 we will have some problems with the tests, they behave different with that version and our current jest version (I checked on the CI). I agree to use Node 10 but we will have to upgrade jest first, our version is slightly old.

PR #59, opened 2019-05-22
Refactor Api Service into common methods

mariosouto:


src/api/apiService.js
```diff
@@ -2,53 +2,73 @@ import { sessionService } from 'redux-react-native-session';
...
-  if (headers.get('access-token')) {
-    const sessionHeaders = {
-      token: headers.get('access-token'),
-      uid: headers.get('uid'),
-      client: headers.get('client')
-    };
-    sessionService.saveSession(sessionHeaders);
-  }
+import handleErrors from './utils/handleErrors';
+import getResponseBody from './utils/getResponseBody';
```
ConstanzaCurzio: Can you put absolute imports here?

src/api/apiService.js
```diff
@@ -2,53 +2,73 @@ import { sessionService } from 'redux-react-native-session';
...
 };
 
-const handleErrors = response =>
-  new Promise((resolve, reject) => {
-    if (!response) {
-      reject({ message: 'No response returned from fetch' });
-      return;
-    }
+class Api {
+  static get(uri, apiUrl = Config.API_URL) {
```
bernabe9: Use `async` / `await` when its possible.

```suggestion
  static async get(uri, apiUrl = Config.API_URL) {
    const requestData = this.buildRequestData(HTTP_VERB.GET);
    await this.loadHeadersAndPerformRequest(uri, apiUrl, requestData);
  }
```

src/api/apiService.js
```diff
@@ -2,53 +2,73 @@ import { sessionService } from 'redux-react-native-session';
...
+    const requestData = this.buildRequestData(HTTP_VERB.GET);
+    return this.loadHeadersAndPerformRequest(uri, apiUrl, requestData);
```
bernabe9: Never use `this` in static methods. `this` reference an instance.

```suggestion
    return Api.loadHeadersAndPerformRequest(uri, apiUrl, requestData);
```

src/api/apiService.js
```diff
@@ -2,53 +2,73 @@ import { sessionService } from 'redux-react-native-session';
...
+
+    if (data) {
+      const decamelizedData = humps.decamelizeKeys(data);
+      requestData.body = JSON.stringify(decamelizedData);
+    }
+
+    return requestData;
+  }
+
+  static loadHeadersAndPerformRequest(uri, apiUrl, data) {
```
bernabe9: Use `async` / `await` and don't modify the input params.

```suggestion
  static async loadHeadersAndPerformRequest(uri, apiUrl, data) {
    const requestData = { ...data }
    try {
      const headers = await Api.getTokenHeader();
      requestData.headers = { ...requestData.headers, ...headers };
    } finally {
      return Api.performRequest(uri, apiUrl, requestData)
    }
  }
```

src/api/apiService.js
```diff
@@ -66,96 +86,13 @@ class Api {
         .then((session) => {
```
bernabe9:
```javascript
static async getTokenHeader() {
  const { token, client, uid } = await sessionService.loadSession();
  return { ACCESS_TOKEN: token, client, uid }
}
```

src/api/apiService.js
```diff
@@ -2,53 +2,73 @@ import { sessionService } from 'redux-react-native-session';
...
 
-const handleErrors = response =>
-  new Promise((resolve, reject) => {
-    if (!response) {
-      reject({ message: 'No response returned from fetch' });
-      return;
-    }
+class Api {
+  static get(uri, apiUrl = Config.API_URL) {
+    const requestData = this.buildRequestData(HTTP_VERB.GET);
```
bernabe9:
```suggestion
    const requestData = Api.buildRequestData(HTTP_VERB.GET);
```

src/api/apiService.js
```diff
@@ -2,53 +2,73 @@ import { sessionService } from 'redux-react-native-session';
...
-const getResponseBody = (response) => {
-  const bodyIsEmpty = response.status === 204;
-  if (bodyIsEmpty) {
-    return Promise.resolve();
+  static patch(uri, data, apiUrl = Config.API_URL) {
+    const requestData = this.buildRequestData(HTTP_VERB.PATCH, data);
+    return this.loadHeadersAndPerformRequest(uri, apiUrl, requestData);
+  }
+
+  static buildRequestData(httpVerb, data = null) {
```
bernabe9:
```suggestion
  static buildRequestData(httpVerb, data = null) {
    return {
      method: httpVerb,
      headers: {
        accept: APPLICATION_JSON,
        [CONTENT_TYPE]: APPLICATION_JSON,
      },
      ...(!!data && { body: JSON.stringify(humps.decamelizeKeys(data)) })
    }
  }
```

peter7z: I think this is a little bit harder to read.
What about:
```suggestion
  static buildRequestData(httpVerb, data = undefined) {
    return {
      method: httpVerb,
      headers: {
        accept: APPLICATION_JSON,
        [CONTENT_TYPE]: APPLICATION_JSON,
      },
      body: data && JSON.stringify(humps.decamelizeKeys(data))
    }
  }
```
I know that `(typeof undefined != typeof null)` but I don't see why that would be a problem.
Or just the `if` version that is already written.

src/api/apiService.js
mariosouto: Actually calling `this` or `Api` in this case is the same. But i see that it could bring some confusion. I'll change it

src/api/apiService.js
juanlacruz: I think it doesn't make sense here since it's on the same folder, wdyt? but I like the idea of using absolute imports 👍

src/api/apiService.js
juanlacruz: Prefer this one @peter7z . I don't see any problems with that, @bernabe9 wdyt?

mariosouto: I went with a mix of both

src/api/apiService.js
mariosouto: I was using finally but i received an error from the lintern. Turns out that finally could led to some unexpected results, so i went with just a `catch`

https://eslint.org/docs/rules/no-unsafe-finally

src/api/apiService.js
```diff
@@ -2,159 +2,83 @@ import { sessionService } from 'redux-react-native-session';
...
+};
 
-    response.json()
-      .then((json) => {
-        const error = json || { message: response.statusText };
-        reject(error);
-      }).catch(() => reject({ message: 'Response not JSON' }));
-  });
+class Api {
+  static async get(uri, apiUrl = Config.API_URL) {
```
bernabe9: you don't need to use `async` if you are not waiting for anything.

bernabe9: In my previous comment, I suggested to use `async` with an await for the `loadHeadersAndPerformRequest`. But if you are just returning the promise it's fine without the `async`

mariosouto: The use of `await` is implicit there. In fact, if i put it i get an error from the lintern

> error  Redundant use of `await` on a return value

bernabe9: My suggestion is to remove the `async`, not to add the `await`. Sorry for the confusion.

PR #60, opened 2019-05-29
Fix navigation options not being set when registering screens

santiestra:
#### Description:
* Copy options into the wrapped component
* Change Component.navigationOptions (not supported in eact-native-navigationsv2) favor of Component.options
---

#### Notes:
Fix based on https://github.com/wix/react-native-navigation/issues/1642#issuecomment-361195519

### DIFF
--- a/src/screens/registerScreen.js
+++ b/src/screens/registerScreen.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import { Provider } from 'react-redux';
 import { Navigation } from 'react-native-navigation';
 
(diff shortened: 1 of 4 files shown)


src/screens/registerScreen.js
```diff
@@ -7,12 +7,20 @@ const registerScreen = (
   Component,
   store,
 ) => {
-  const ConnectedComponent = props =>
-    <Provider store={store}>
-      <Component
-        {...props}
-      />
-    </Provider>;
+  class ConnectedComponent extends PureComponent {
```
bernabe9: Why not keep it as a function component?

```suggestion
  const ConnectedComponent = props =>
    <Provider store={store}>
      <Component {...props} />
    </Provider>;
  ConnectedComponent.options = Component.options;
```

PR #62, opened 2019-05-30
Remove unnecessary call to start when initializing the application

mariosouto:


src/App.js
```diff
@@ -23,17 +23,19 @@ class App {
     const session = store.getState().get('session');
     const isAuthenticated = session.get('authenticated');
     const user = session.get('user');
-    const shouldUpdate = this.isAuthenticated !== isAuthenticated && (!isAuthenticated || !user.isEmpty());
+
+    const userLoggedOut = this.isAuthenticated !== isAuthenticated && !isAuthenticated;
+    const userLoggedIn = this.isAuthenticated !== isAuthenticated && !user.isEmpty();
```
santiestra: what do you think about also moving `this.isAuthenticated !== isAuthenticated` to a const, for example `authenticationChanged`?

mariosouto: Sounds good, will do

PR #63, opened 2019-05-30
Add context and HOC componentId/navigation

peter7z:
This PR wraps registered screens in a context provider with componentId as value.
Then it implements a HOC that consumes this Context and injects a navigation prop with componentId already placed as the first argument of some Navigation functions.

What do you think?

### DIFF
--- a/src/containers/LoginScreen/index.js
+++ b/src/containers/LoginScreen/index.js
@@ -1,8 +1,7 @@
 import React from 'react';
 import { Text, View, Button } from 'react-native';
 import { connect } from 'react-redux';
-import { Navigation } from 'react-native-navigation';
-import { func, string } from 'prop-types';
+import { func, object } from 'prop-types';
 
 import LoginForm from 'components/user/LoginForm';
 import { login } from 'actions/userActions';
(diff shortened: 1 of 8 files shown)


src/containers/LoginScreen/index.js
```diff
@@ -40,4 +40,4 @@ const mapDispatch = dispatch => ({
   login: user => dispatch(login(user))
 });
 
-export default connect(null, mapDispatch)(LoginScreen);
+export default connect(null, mapDispatch)(withNavigation(LoginScreen));
```
felipegarcia92: Wouldn't you think that we could place these `withNavigation` calls in `src/screens/index.js` instead of the container? 

Maybe it doesn't really make a difference, thought of that to not keep wrapping containers here, and thinking that most of registered screens there will need it, if not all.

peter7z: Will do.
Regarding your other question, i thought we could user this context for more that just navigation, that's why i called it screenContext.

agustito37:
Nice, I like this.

peter7z:
Shall we merge this?

felipegarcia92:
@peter7z have the honors

PR #65, opened 2019-06-04
Replace immutable with immer

peter7z:
For a description and discussion go to [this PR](https://github.com/rootstrap/react-redux-base/pull/118/files).

### DIFF
--- a/src/selectors/sessionSelector.js
+++ b/src/selectors/sessionSelector.js
@@ -1,6 +1,0 @@
-import { createSelector } from 'reselect';
-
-export const getUser = createSelector(
-  state => state.getIn(['session', 'user']),
-  user => user.toJS()
-);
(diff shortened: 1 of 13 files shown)


src/selectors/sessionSelector.js
```diff
@@ -1,6 +1,6 @@
 import { createSelector } from 'reselect';
 
 export const getUser = createSelector(
```
bernabe9: This selector is useless now :(

The goal of the selectors is to avoid recomputing data unless is necessary. In this case, we were avoiding to compute the function `.toJS()` that is expensive. But now, we are just accessing an object's property. It's more expensive the selector itself than the function to get the data.

peter7z: We could keep them around to calculate derived data, what do you think?
Or maybe it would be best to leave that to each project.

PR #67, opened 2019-06-07
Replace promises for async/await

mariosouto:


src/api/utils/getResponseBody.js
```diff
@@ -1,9 +1,4 @@
-export default (response) => {
-  const isBodyEmpty = response.status === 204;
-
-  if (isBodyEmpty) {
-    return Promise.resolve();
-  }
-
-  return response.json();
+export default async (response) => {
```
ConstanzaCurzio: This is async but has no await inside 🤔

src/api/utils/handleErrors.js
```diff
@@ -1,29 +1,26 @@
...
       });
+  } catch (e) {} // eslint-disable-line
 
-    response.json()
-      .then((json) => {
-        const error = json || { message: response.statusText };
-        reject(error);
-      }).catch(() => reject({ message: 'Response not JSON' }));
-  });
+  throw await response.json()
```
mcousillas6: instead of mixing await and .then syntax, why don't you go with something like a helper to parse the error
```javascript
const parseError = async (response) =>  {
try {
const json = await response.json();
return json || { message: response.statusText };
} catch (error) {
return { message: 'Response not JSON };
}
}
```
and then just call `throw new Error(await parseError(response));`

src/api/utils/handleErrors.js
```diff
@@ -1,29 +1,26 @@
...
 
-    sessionService.loadSession()
+  try {
+    await sessionService.loadSession()
       .then(() => {
         if (response.status === 401) {
           sessionService.deleteSession();
         }
       });
+  } catch (e) {} // eslint-disable-line
```
mcousillas6: maybe add some logging here? so we don't lose the error completely

src/api/utils/handleErrors.js
```diff
@@ -1,29 +1,26 @@
...
-      return;
-    }
+  if (response.ok) {
+    await saveSessionHeaders(response.headers);
+    return response;
+  }
 
-    sessionService.loadSession()
+  try {
+    await sessionService.loadSession()
```
mcousillas6: Why mix await and .then syntax and not just do: 
```javascript 
await sessionService.loadSession()
if (response.status === 401) {
    sessionService.deleteSession();
}
```

src/api/utils/handleErrors.js
bernabe9: @mariosouto can you do the same with the react base?

src/api/utils/getResponseBody.js
mariosouto: It does, but it's implicit in the `response.json()`. The thing is that `response.json()` returns a promise, so for consistency i choose to return an empty Promise in the other cases, like we did before the change.

What do you think? Also including @mcousillas6 @bernabe9 on this

agustito37: It's weird using `camelizeKeys` if the `response` could be `undefined`. I would rather `await` for the `response.json()` to finish in order to camelize its keys.
if (!isBodyEmpty) return response.json();
`if (!isBodyEmpty) return humps.camelizeKeys(await response.json());`

PR #69, opened 2019-06-10
Remove call to isEmpty that was left from migration to immer

santiestra:
#### Description

isEmpty is a function from immutable Map
Change it for js check for emptyness
Update README

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import { Navigation } from 'react-native-navigation';
 import { sessionService } from 'redux-react-native-session';
+import isEmpty from 'lodash/isEmpty';
 
 import configureStore from './store/configureStore';
 import registerScreens from './screens';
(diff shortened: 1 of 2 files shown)


peter7z:
Good catch. Thank you!

santiestra:
> Good catch. Thank you!

@peter7z Thank you for removing immutable!

src/App.js
```diff
@@ -25,7 +25,7 @@ class App {
 
     const authenticationChanged = this.isAuthenticated !== isAuthenticated;
     const userLoggedOut = authenticationChanged && !isAuthenticated;
-    const userLoggedIn = authenticationChanged && !user.isEmpty();
+    const userLoggedIn = authenticationChanged && Object.keys(user).length > 0;
```
bernabe9: is it possible for `user` to be `undefined`? If yes, this will fail, otherwise looks good 👍

peter7z: I used this:
```
export const isEmpty = (obj) => {
  for(var key in obj) {
      if(obj.hasOwnProperty(key))
          return false;
  }
  return true;
};
```

santiestra: Thanks for the review :+1:!. 
It seems it's always defined, otherwise, `user.isEmpty()` would've also failed. 
I think that the initial state of the session reducer from the session library has an empty object for the user.

mato2593: Since we are using lodash, why not using lodash's `isEmpty`?

santiestra: I didn't know the project had lodash! Thanks!

PR #72, opened 2019-06-17
Implement react hooks

aguscha333:
## Implement react hooks

This PR implements hooks in the native base, this is a port of the work done by @felipegarcia92 in the web base.

### Tested on
- [x] iOS
- [x] Android


src/containers/MainScreen/index.js
```diff
@@ -1,25 +1,29 @@
...
-  <View style={styles.container}>
-    <Text>Hey {email}, you&#39;re logged in!</Text>
-    <Button
-      onPress={logout}
-      title={translate('MAIN_SCREEN.logout')}
-    />
-  </View>
-);
+const MainScreen = () => {
+  const dispatch = useDispatch();
```
agustito37: A small detail. Can you add a line as you do in the signup screen?

peter7z:
Great job, thanks! It's awesome to have this in native too.
As a side note:
The app class is not a React component and it's in the file where the store is defined. I am not sure we should add hooks there, they might not even work outside of a component.

mcousillas6:
> Great job, thanks! It's awesome to have this in native too.
> As a side note:
> The app class is not a React component and it's in the file where the store is defined. I am not sure we should add hooks there, they might not even work outside of a component.

Yes, if the App class is not a react component already, we don't need to migrate it, since hooks only work for functional components.

src/containers/MainScreen/index.js
aguscha333: fixed!

PR #74, opened 2019-06-18
Status reducer

mcousillas6:
Added a generic reducer that track the status of every action dispatched. The main goal is to reduce boilerplate and standardise status and error handling in one place.

The status reducer matches against `ACTION_(REQUEST|ERROR|SUCCESS|RESET)` to define the status of the action between `STARTED|NOT_STARTED|LOADING|ERROR`.
If the action has an error key, it will be stored alongside the status.
Added hooks to query the store.
This way, we only need to dispatch the correct action types, and we are good to go, the statusReducer takes care of the rest, and you can use the provided hooks `useLoading` `useError` and `useStatus` to query that state for each action.


agustito37:
Cool, I missed the _RESET section. I was thinking, what if we can store the value of the response too? You would have all the necessary things to manage a request in one place.

mcousillas6:
> Cool, I missed the _RESET section. I was thinking, what if we can store the value of the response too? You would have all the necessary things to manage a request in one place. This is not always appropriate but It would be nice. It wouldn't be a status reducer it would be a request reducer. I am thinking if this is usable. Maybe not XD.

Yes, I think that would end up getting out of hand a bit too quick 😆 .
Status is something that is consistent and doesn't rely on the structure of any external dependency (The API can behave in any way and this will still work for example) that's why I stayed away from modeling responses.

felipegarcia92:
@mcousillas6 can you post an example use case of this? Sorry, I'm still trying to wrap my head around this.

mcousillas6:
> @mcousillas6 can you post an example use case of this? Sorry, I'm still trying to wrap my head around it.

I've already used it on the login flow to display the loading indicator.

If you check the diffs, now i'm dispatching a new action `LOGIN_REQUEST` before awaiting for the `apiService.login` response.
- That `LOGIN_REQUEST` is what updates the status reducer's value to `LOADING`.
- After the response arrives, the `LOGIN_SUCCESS` is dispatched (same as before), but now the statusReducer catches that action and updates the status again to `SUCCESS`.

Finally ,In the loginContainer i'm using the `userLoading` selector with the value `'LOGIN` to watch the value of actionStatus[LOGIN] to pass it as the `isSubmitting` prop

Not sure if that makes the flow clear, let me know if you need another example 😄

felipegarcia92:
Cool, I think I gotcha. To confirm, let's take a simple target list example, I'd end up with a component like this one:

```
const TargetList = ({ props }) => {
  const { status, error } = useStatus('TARGET')
  const loading = useLoading('TARGET')
  // and you need some other way to fetch the targets themselves

  return (
     ... use loading here, use error here, use targets here, use status
  )
}
```

Is that correct? It does help with some code.

Before seeing this, and having added a `useSession` hook myself, in my head I was thinking hooks on an per-reducer/entity basis. Something like:

```
const TargetList = ({ props }) => {
  const { loading, error, targets } = useTargets()

  return (
     ... use loading here, use error here, use targets here
  )
}
```

and have `useTargets()` return the object with those keys using `useSelector` and selectors, or whatever suited. But that's another idea, we can discuss variations of course.

Also not sure how the status itself can help. When fetching async data we normally have the loading, success (where you have a value stored) and error statuses, which are already covered by the `loading` and `error` keys.

src/hooks/useError.js
```diff
@@ -0,0 +1,8 @@
+import { useSelector } from 'react-redux';
+
+const useError = action => useSelector(({ actionStatus }) => {
```
felipegarcia92: I can get the error already using `useStatus`, right? maybe it's redundant

mcousillas6: You can, but it's a bit more verbose, i left the `useLoading` and `useError` implemented "specializations" of the `useStatus`, since many times you just want to know if something is actually loading, or display an error if present. The `useStatus` definitely covers both cases, but takes a bit more code to do so.

mcousillas6:
> Cool, I think I gotcha. To confirm, let's take a simple target list example, I'd end up with a component like this one:
> 
> ```
> const TargetList = ({ props }) => {
>   const { status, error } = useStatus('TARGET')
>   const loading = useLoading('TARGET')
>   // and you need some other way to fetch the targets themselves
> 
>   return (
>      ... use loading here, use error here, use targets here, use status
>   )
> }
> ```
> 
> Is that correct? It does help with some code.
> Also not sure how the status itself can help. When fetching async data we normally have the loading, success (where you have a value stored) and error statuses, which are already covered by the `loading` and `error` keys.

The status itself is an enum that describes all the possible states of an action (because it can be in more states than loading/success, for example: not started if you haven't dispatched the action).
This comes in handy when you want a more fine grained control of such status, for example, navigate to a screen on success, or only dispatching the action if it hasn't already been, etc.

I think the confusion between status and loading/error came from the `useLoading` and `useError` selectors, if you check them, they use the status key under the hood, but maching against an specific case.
So, `useLoading` is the same as calling `useStatus` and then doing `status === LOADING`

> 
> Before seeing this, and having added a `useSession` hook myself, in my head I was thinking hooks on an per-reducer/entity basis. Something like:
> 
> ```
> const TargetList = ({ props }) => {
>   const { loading, error, targets } = useTargets()
> 
>   return (
>      ... use loading here, use error here, use targets here
>   )
> }
> ```
> 
> and have `useTargets()` return the object with those keys using `useSelector` and selectors, or whatever suited. But that's another idea, we can discuss variations of course.
> 

Yeah, I agree with that approach, the thing is, you could have more than one action associated with an entity, so having a hook that fetches the status and the entity. 
For example: session has Login, logout and Sign in actions associated, but they all work on the same entity. I think it's clearer if you have separation between the status of those 3 actions and your entity. 
It's definitely a topic to review and discuss though, since there is no clear answer.

mcousillas6:
> Nice initiative 👍 Maybe I'm not getting something but do we still need to define flow events like (LOGIN_SUCCESS, ORDER_SUCCESS, ...) in order to identify which events were triggered, right?
> 
> So correct me if I'm wrong, you're wrapping the state in order to always have the same status name (REQUEST, SUCCESS, ERROR...) right?

Yep, that's correct. At least for now, i've seen some people implementing similar things, and also action creators that handle the triggers, so like a dispatchRequest('LOGIN') that dispatches 'LOGIN_REQUEST' and so son.

We could add something similar, but i don't know if it would make things a bit obscure

src/containers/LoginScreen/index.js
```diff
@@ -15,12 +16,14 @@ const LoginScreen = ({ navigation }) => {
...
   );
+  const isLoading = useLoading('LOGIN');
+
   return (
     <View style={styles.container}>
       <Text style={styles.welcome}>
         {translate('SIGN_IN.title')}
       </Text>
-      <LoginForm onSubmit={loginRequest} />
+      <LoginForm onSubmit={loginRequest} submitting={isLoading} />
```
bernabe9: You don't need to add the property `submitting` because `redux-form` is already adding it

felipegarcia92:
> The status itself is an enum that describes all the possible states of an action (because it can be in more states than loading/success, for example: not started if you haven't dispatched the action).
> This comes in handy when you want a more fine grained control of such status, for example, navigate to a screen on success, or only dispatching the action if it hasn't already been, etc.
> 
> I think the confusion between status and loading/error came from the `useLoading` and `useError` selectors, if you check them, they use the status key under the hood, but maching against an specific case.

Can you explain the `not started` case? If you haven't dispatched the action then you wouldn't have neither the value, nor loading variable, nor an error. I think all of them are deducible from the three variables we hold now: `loading`, `error` and the value itself.

> So, useLoading is the same as calling `useStatus` and then doing `status === LOADING`

Regarding `useLoading`, `useError`, and `useStatus`, for me it's unnecessary code in the base that's unused. Will we have a `useNotStarted` too? For me that's creating too many specific hooks when you can just compare.

> Yeah, I agree with that approach, the thing is, you could have more than one action associated with an entity, so having a hook that fetches the status and the entity.
> For example: session has Login, logout and Sign in actions associated, but they all work on the same entity. I think it's clearer if you have separation between the status of those 3 actions and your entity.
> It's definitely a topic to review and discuss though, since there is no clear answer.

I get your point, but in practice, normally one of the following is the case:

1. They are split in separate pages/components so they don't collide. I even think we **should** encapsulate it that way. 

2. You're ultimately interested in the result itself, even though the actions are separate/different. Using the session example, you care about having the session, not having it, or being loading it. But don't care about which action triggered that. Going back to Target examples, if you have a list of targets and there's an action for fetching them and another one for updating them, you're interested in the most up-to-date targets list, and if you want to know if the update call is loading, you can just use the typical `loading` variable in the reducer or if there's a conflict between them running at the same time (rare), a new one called, let's say, `loadingUpdate`. But I don't think you'll need it.

Also, you can always have two or more loading variables which you can also return in the `useWhatever` hook if you have one.

This is just my opinion, good work btw! Of course I'd like more opinions on this. Maybe it's beneficial to have and I'm not seeing it, but I prefer not to add extra/auxiliar code to the base if not needed and is not tech debt.

mcousillas6:
> Can you explain the not started case? If you haven't dispatched the action then you wouldn't have neither the value, nor loading variable, nor an error. I think all of them are deducible from the three variables we hold now: loading, error and the value itself.

The not started case is useful as a pattern to avoid awaiting for dispatched actions on your components, instead, you can listen to status changes and, for example, dispatch navigation/reset/etc actions on specific states. 

>I get your point, but in practice, normally one of the following is the case:

>They are split in separate pages/components so they don't collide. I even think we should encapsulate it that way.

>You're ultimately interested in the result itself, even though the actions are separate/different. Using the session example, you care about having the session, not having it, or being loading it. But don't care about which action triggered that. Going back to Target examples, if you have a list of targets and there's an action for fetching them and another one for updating them, you're interested in the most up-to-date targets list, and if you want to know if the update call is loading, you can just use the typical loading variable in the reducer or if there's a conflict between them running at the same time (rare), a new one called, let's say, loadingUpdate. But I don't think you'll need it.

>Also, you can always have two or more loading variables which you can also return in the useWhatever hook if you have one.

>This is just my opinion, good work btw! Of course I'd like more opinions on this. Maybe it's beneficial to have and I'm not seeing it, but I prefer not to add extra/auxiliar code to the base if not needed and is not tech debt.

Yeah, that's a totally valid approach, I think both work. Maybe we could come up with some generic way of implementing your alternative to see if we could end up with something that saves us boilerplate as well.

src/reducers/statusReducer.js
```diff
@@ -0,0 +1,38 @@
+import { produce } from 'immer';
+import { NOT_STARTED, LOADING, SUCCESS, ERROR } from 'constants/status';
+
+const handleAction = (state, action) => {
+  const { type, error } = action;
+
+  const matchesStart = /(.*)_(REQUEST)/.exec(type);
```
peter7z:
```suggestion
  const matchesStart = /(.*)_REQUEST/.exec(type);
```

PR #76, opened 2019-07-02
Updated react native to 0.59.10

mcousillas6:
https://github.com/facebook/react-native/releases/tag/v0.59.10

### DIFF
--- a/android/app/src/main/java/com/reactnativebase/MainApplication.java
+++ b/android/app/src/main/java/com/reactnativebase/MainApplication.java
@@ -3,6 +3,7 @@ package com.reactnativebase;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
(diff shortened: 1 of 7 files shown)


tamarasuarezrod:
we should use the version 1.0.6 for redux-react-native-session package to handle the change of async storage from react-native to react-native-community

android/app/src/main/java/com/reactnativebase/MainApplication.java
```diff
@@ -3,6 +3,7 @@
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.reactnativecommunity.asyncstorage.AsyncStoragePackage;
```
agustito37: You need to instanciate this package. 
https://github.com/react-native-community/async-storage/blob/LEGACY/docs/Linking.md
We should also check if another imported package is not being used.

PR #79, opened 2019-07-08
Prettier integration and refactor

nsantos16:
## Description

Based on  [PR](https://github.com/rootstrap/eye-level-flex-react/pull/5) made by @aguscha333 

## Pros
- Easy to adopt
- Helping Newcomers
- Building and enforcing a style guide

## Cons
- In some cases it's better unformatted code

## Documentation
- [Prettier docs](https://prettier.io/docs/en/install.html)
- [Precommit with prettier](https://prettier.io/docs/en/precommit.html)
- [Formatting Code Automatically](https://facebook.github.io/create-react-app/docs/setting-up-your-editor#formatting-code-automatically)
## Changes

- Added prettier
- Added hook to commits with husky
- Added lint-staged
- Configuration of `.eslintrc` and `.prettierrc`

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "printWidth": 100,
+  "singleQuote": true
+}
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -52,6 +52,8 @@ checks:
 plugins:
   eslint:
     enabled: true
+    config:
+      config: ~/.eslintrc
 exclude_patterns:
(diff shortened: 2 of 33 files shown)


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "printWidth": 100,
```
aguscha333: same comment as in the other PR, we need to decide if this is the preferred print width

PR #80, opened 2019-07-16
React-navigation integration

mcousillas6:
### Migrated from React-Native-Navigation to [React-Navigation](https://reactnavigation.org/)

React navigation has a few advantages over RNNav:

- It's written on JS (with a few native primitives for performance), so it's easier to debug.
- Animations are done using the `Animatable` api.
- The Navigator API is way more clear and straightforward than RNNav v2 (which changed a lot from v1).
- Everything is a react component.
- It's the community standard, so we would have access to better documentation, faster updates, and a bigger community using it.
- No weird gradle configs on android 🎉


src/containers/SignUpScreen/index.js
```diff
@@ -2,36 +2,45 @@ import React, { memo, useCallback } from 'react';
...
 
+  useSessionChangeEffect(
+    user => {
+      const userLoggedIn = !isEmpty(user);
+      if (userLoggedIn) {
+        navigation.navigate(MAIN_SCREEN);
+      }
+    },
+    [navigation]
+  );
```
peter7z: You could abstract these 3 hook implementations (Login and MainScreen) into their own hook.

src/containers/SignUpScreen/index.js
```diff
@@ -2,36 +2,45 @@ import React, { memo, useCallback } from 'react';
...
+
   return (
     <View style={styles.container}>
       <Text style={styles.welcome}>{translate('SIGN_UP.title')}</Text>
       <SignUpForm onSubmit={signUpRequest} />
-      <Button title={translate('SIGN_IN.title')} onPress={navigation.pop} />
+      <Button title={translate('SIGN_IN.title')} onPress={() => navigation.goBack()} />
```
peter7z:
```suggestion
      <Button title={translate('SIGN_IN.title')} onPress={navigation.goBack} />
```

android/app/src/main/java/com/reactnativebase/MainApplication.java
```diff
@@ -3,52 +3,54 @@
...
+    @Override
+    public boolean getUseDeveloperSupport() {
+      return BuildConfig.DEBUG;
+    }
+
+    @Override
+    protected List<ReactPackage> getPackages() {
+      return Arrays.<ReactPackage>asList(
+          new MainReactPackage(),
+            new RNGestureHandlerPackage(),
```
agustito37:
```suggestion
          new RNGestureHandlerPackage(),
```

src/App.js
```diff
@@ -1,62 +1,20 @@
...
-
-class App {
-  constructor() {
-    this.isRunning = false;
-    this.isAuthenticated = false;
 
+export default function App() {
+  useEffect(() => {
     sessionService.initSessionService(store);
+  }, []);
```
agustito37: Why the empty array?

src/containers/AppLoader/index.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { View, ActivityIndicator, StyleSheet } from 'react-native';
+import { object } from 'prop-types';
+import isEmpty from 'lodash/isEmpty';
+import useSessionChangeEffect from 'hooks/useSessionChangeEffect';
+import { LOGIN_SCREEN, MAIN_SCREEN } from 'constants/screens';
+
+const styles = StyleSheet.create({
```
agustito37: Should this style be here?

src/containers/LoginScreen/index.js
```diff
@@ -2,38 +2,44 @@ import React, { memo, useCallback } from 'react';
...
   const loginRequest = useCallback(user => dispatch(login(user)), [dispatch]);
+
+  useSessionChangeEffect(
+    user => {
+      const userLoggedIn = !isEmpty(user);
+      if (userLoggedIn) {
+        navigation.navigate(MAIN_SCREEN);
+      }
+    },
+    [navigation]
```
agustito37: I'm trying to ### understand this, why are you triggering the hook when the navigation object changes?

src/containers/AppLoader/index.js
mato2593: I assumed he did this because this is not the purpose of the draft, but if we're gonna add this to the base it should be in a styles file of course

src/App.js
agustito37: I found why :) nevermind

src/containers/AppLoader/index.js
```diff
@@ -0,0 +1,39 @@
...
+        navigation.navigate(MAIN_SCREEN);
+      } else {
+        navigation.navigate(LOGIN_SCREEN);
+      }
+    },
+    [navigation]
+  );
+
+  return (
+    <View style={styles.container}>
```
bernabe9: Is this going to show a loading indicator when is checking the session?

I think we should show the splash screen until the app is ready to launch. Otherwise, this screen will look like a flickering screen (splash screen => loading screen for a second => user screen). WDYT?

src/navigators/index.js
```diff
@@ -0,0 +1,27 @@
+import { createStackNavigator, createAppContainer, createSwitchNavigator } from 'react-navigation';
+import LoginScreen from 'containers/LoginScreen';
+import MainScreen from 'containers/MainScreen';
+import SignUpScreen from 'containers/SignUpScreen';
+import AppLoader from 'containers/AppLoader';
+
+const AuthStack = createStackNavigator({
```
bernabe9: I like these stacks 👍

src/containers/LoginScreen/index.js
mariosouto: I'm not getting this too 🤔

src/containers/MainScreen/index.js
```diff
@@ -23,12 +36,19 @@ const MainScreen = () => {
...
 };
 
-MainScreen.options = {
-  topBar: {
-    title: {
-      text: translate('MAIN_SCREEN.title')
-    }
+MainScreen.navigationOptions = {
+  title: translate('MAIN_SCREEN.title'),
+  headerStyle: {
```
mariosouto: If i recall you can define these styles when you define the stack, maybe it's better to all this in the stack definition? WDYT?

src/containers/LoginScreen/index.js
mcousillas6: If I remember correctly, it shouldn't, I added it just as a good practice (since all props used inside an effect should be  passed as dependencies). But we can definitely omit it.

src/containers/AppLoader/index.js
mcousillas6: Yes, I think that would look better, we would need to come up with some kind of "fake" splash though, since RNNav needs to display something and does not have control over the current splash.

src/App.js
```diff
@@ -1,62 +1,20 @@
...
 
 const store = configureStore({});
-registerScreens(store);
-
-class App {
-  constructor() {
-    this.isRunning = false;
-    this.isAuthenticated = false;
 
+export default function App() {
```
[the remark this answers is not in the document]
mcousillas6: using these because functions are cheaper than arrow functions.

src/containers/MainScreen/index.js
mcousillas6: Yeah, if there are global stacks, we should use the `defaultNavigationOptions` on the stack itself. In this case I just wanted to change a bit the styles of the main screen to leave an example on how to pass `navigationOptions` to a specific screen

src/containers/SignUpScreen/index.js
mcousillas6: Yep, I like the idea

src/containers/SignUpScreen/index.js
mcousillas6: Tried it and it does not work, not sure why though 🤔.

src/App.js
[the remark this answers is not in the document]
bernabe9: 🤔

src/containers/AppLoader/index.js
bernabe9: Isn't possible to use `react-native-splash-screen` to programmatically hide the splash screen?

src/App.js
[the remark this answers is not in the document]
mcousillas6: NVM, did some more digging and they are almost the same. 
The only gotcha of arrows is to avoid exporting unnamed functions, because you won't have the `displayName` (mostly for debugging)

src/containers/AppLoader/index.js
mcousillas6: https://github.com/crazycodeboy/react-native-splash-screen this one? Looks like we could, my only concern would be RN 0.60 support, last update was on February 🤔

src/containers/SignUpScreen/index.js
agustito37: Without the arrow function, the function goBack is probably executing in the event handler's context and not in the navigation's context. But who knows 🤔

PR #81, opened 2019-07-23
RN 0.60.4 migration

mcousillas6:
## Do not merge RNNav alpha fails on android for now

Migrated RN to latest 0.60.4 release

### Notes:
- `react-native-navigation` only supports rn0.60+ on it's v3, which is currently on alpha stage, it seems pretty stable and does not have breaking changes, so we could wok with this one until the stable release (or migrate to react-navigation).

It's probably best if we keep this a separate branch until all dependencies are fully stable.


android/settings.gradle
```diff
@@ -1,13 +1,9 @@
 rootProject.name = 'ReactNativeBase'
-include ':@react-native-community_async-storage'
-project(':@react-native-community_async-storage').projectDir = new File(rootProject.projectDir, '../node_modules/@react-native-community/async-storage/android')
```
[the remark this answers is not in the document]
mcousillas6: It gets auto linked

android/app/build.gradle
```diff
@@ -127,63 +152,64 @@ android {
         }
     }
     signingConfigs {
-        release {
```
agustito37: Shouldn't we keep the release configuration?

android/app/build.gradle
```diff
@@ -127,63 +152,64 @@ android {
...
         }
     }
     buildTypes {
+        debug {
+            signingConfig signingConfigs.debug
+        }
         release {
+            // Caution! In production, you need to generate your own keystore file.
+            // see https://facebook.github.io/react-native/docs/signed-apk-android.
+            signingConfig signingConfigs.debug
```
agustito37:
```suggestion
            signingConfig signingConfigs.release
```
?

android/app/build.gradle
mcousillas6: yes, good catch!

src/App.js
```diff
@@ -19,6 +19,15 @@ class App {
     store.subscribe(this.onStoreUpdate.bind(this));
   }
 
+  setDefaultOptions() {
+    Navigation.setDefaultOptions({
+      layout: {
+        backgroundColor: '#FAFAFA',
```
agustito37: Shouldn't this be in a constant file or something?

.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "printWidth": 100,
-  "singleQuote": true
-}
\ No newline at end of file
+  "singleQuote": true,
+  "trailingComma": "es5"
```
aguscha333: why this change?

mcousillas6: Because it was really inconsistent, test files and other configs all had trailing commas, and the ones that prettier handled, did not. Also, it [has some advantages](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8) WDYT?

peter7z: Go team trailing commas!

ios/Podfile
```diff
@@ -0,0 +1,69 @@
...
+
+  target 'ReactNativeBaseTests' do
+    inherit! :search_paths
+    # Pods for testing
+  end
+
+  use_native_modules!
+end
+
+target 'ReactNativeBase-tvOS' do
```
mariosouto: I don't know iOS but is this necessary?

mcousillas6: It comes by default with react native, we could remove it, but it makes no harm and we could end up breaking some automatic stuff from rn 🤷‍♂

PR #82, opened 2019-07-25
Feature/ios build targets

mcousillas6:
## TL;DR
Added different build targets to the iOS project.
With this we can:
- Have three separate environments to match the backend.
- No need to change configs every time.
- You can have all three apps installed on your device, or on testflight.
- Ability to have separate app-icons, splash screens, etc.
- Separate certificates and APNs setup (if needed).
- If you don't need them, you can just ignore them (or delete them altogether), they won't conflict with the default RN target.

## Notes:
- The only "trick" here is on the podfile, now there are three targets, one for each xcode target. In the case that you need to link native modules manually **this is only for modules that not support RN 0.60 autolink feature** you will need to update `Staging` and `Develop` targets by hand to match the default target.

- The PR is targeting the `feature/RN-60` branch because i'm relying on some of RN new features (such as autolink).


ios/Podfile
```diff
@@ -42,6 +45,28 @@ target 'ReactNativeBase' do
   use_native_modules!
 end
 
+target 'ReactNativeBase-Staging' do
+    # Pods for ReactNativeBase
+    base_pods
+
+    pod 'react-native-config', :path => '../node_modules/react-native-config'
```
agustito37: Why these pods are not in base pods if we are repating them?

mcousillas6: I just moved all react-native base dependencies to `base_pods` wasn't sure about these three since they are external dependencies. But yeah, we can move them. WDYT?

agustito37: Sure; I am thinking a case where we could differentiate pods by environment, shouldn't be  a common thing I think, but having the definition in the base could help if happens.

mcousillas6: Yeah, the only scenario that i could think of is having some native logging tool (like crashlytics) that you only want to setup for the prod target 🤔

PR #83, opened 2019-08-16
React-Native 0.60.5 upgrade

mcousillas6:
diff from https://react-native-community.github.io/upgrade-helper/?from=0.60.4&to=0.60.5


mcousillas6:
> Sure, why not...
> We should investigate what's happening with the tests..

I'll check what's happening, they are running just fine locally.

PR #84, opened 2019-08-19
Refactor & cleanup after hooks migration

mcousillas6:
Did some cleanup and refactoring of some stuff that was old/needed to change after migrating to use hooks:

- Removed duplicate key from prettier config that was left from the migration to rn 0.60.5.
- Fixed some prettier issues that were still pending.
- Renamed`containers` to `screens` since we won't have containers anymore.
- Renamed `userApi` to `userService` to keep naming consistent 
- Created a separate `services` folder to keep things separate with the base implementation of the api.
- Removed the global `actionTypes` file, so we start separating actions by resource (so it does not get out of hand when we have 100 actions in one file). 
- Fixed a navigation loop issue on login.

Note: Tests are still failing due to Enzyme's `mount` not working, still haven't found out why


jsconfig.json
```diff
@@ -2,20 +2,48 @@
...
-      "containers/*": ["src/containers/*"],
-      "hooks/*": ["src/hooks/*"],
-      "locale/*": ["src/locale/*"],
-      "reducers/*": ["src/reducers/*"],
-      "screens/*": ["src/screens/*"],
-      "selectors/*": ["src/selectors/*"],
-      "store/*": ["src/store/*"],
-      "utils/*": ["src/utils/*"],
-      "navigators/*": ["src/navigators/*"]
+      "*": [
```
peter7z: What rule is forcing this change? I like the older version better.

mcousillas6: Just ran prettier auto fix, not sure

src/App.js
```diff
@@ -10,7 +10,8 @@ const store = configureStore({});
 export default function App() {
   useEffect(() => {
     const initSession = async () => {
-      await sessionService.initSessionService(store);
+      sessionService.initSessionService(store);
+      await sessionService.refreshFromLocalStorage();
```
mato2593: `initSessionService` calls `refreshFromLocalStorage`, so this is not needed

src/screens/LoginScreen/index.js
```diff
@@ -3,24 +3,29 @@ import { useDispatch } from 'react-redux';
...
-  const loginRequest = useCallback(user => dispatch(login(user)), [dispatch]);
+  const loginRequest = useCallback(
+    async user => {
+      await dispatch(login(user));
+      navigation.navigate(MAIN_SCREEN);
+    },
+    [dispatch],
+  );
 
-  useNavigateOnLoginEffect(navigation);
```
mato2593: Why did you remove the use of this hook?

mcousillas6: I was getting some really weird loops when logging in due to how the section actions are dispatched.

src/screens/LoginScreen/index.js
```diff
@@ -3,24 +3,32 @@ import { useDispatch } from 'react-redux';
...
+import useNavigateOnLoginEffect from 'hooks/useNavigateOnLoginEffect';
 import styles from './styles';
 
 const LoginScreen = memo(({ navigation }) => {
   const dispatch = useDispatch();
-  const loginRequest = useCallback(user => dispatch(login(user)), [dispatch]);
+  const loginRequest = useCallback(
+    async user => {
+      await dispatch(login(user));
+      // navigation.navigate(MAIN_SCREEN);
```
mariosouto: Is this comment necessary?

mcousillas6: Nope, good catch

PR #85, opened 2019-08-20
ApiService Refactor

mcousillas6:
Refactored the `api` module too add support for interceptors and cleaned up a bit of the architecture of the whole thing. 

If you guys like the change, I'll add extra test coverage and JSdocs to keep things tidy.

### Changes

- ApiService now exports a singleton instance with default configs. This helps to remove repetition (passing the baseUrl on every api call for example) and lets us store config stuff like headers, interceptors, etc.

- Cleaned up the whole parse/process body and errors to make a it a bit more clear.

- `processResponse` now injects a `body` property to fetch's response (both on error and ok cases), this should make things a bit more flexible since we will always have all the response information to work with (eg: use the status code to dispatch some action). So instead of just using response as your JSON payload, you should do response.body, the rest of the behaviour is unchanged.

- Removed the `getRequestHeader` method, the default headers can be ser via the constructor, and if any conditional header is needed, requestInterceptors are the way to go.

- Added `requestInterceptors` and `responseInterceptors` that work the same way as axios do. You can call `requestInterceptors/responseInterceptors.use(fullfill, failure)` each callback receives either the request config/response con the fullfill case and the error/errorResponse on the failure case.


peter7z:
Why not just axios it all out?

mcousillas6:
> Why not just axios it all out?

We talked about it on slack a while ago, and fetch had some advantages. So if we weren't going to change packges, we might as well add the missing features 🤷‍♂

src/api/utils/InterceptorManager.js
```diff
@@ -0,0 +1,38 @@
...
+  handlers = [];
+
+  use(fullfill, failure) {
+    this.handlers.push({
+      fullfill,
+      failure,
+    });
+  }
+
+  remove(index) {
```
mariosouto: Not sure if we need this 🤔 , in which scenario do you imagine using it?

src/api/utils/InterceptorManager.js
```diff
@@ -0,0 +1,38 @@
+class InterceptorManager {
+  handlers = [];
+
+  use(fullfill, failure) {
```
mariosouto: I'm not sure if i'm understanding this correctly but do we need to call `use` this way? I don't think you are using `failure` right now.

Is it an error or is it because you didn't believe it was worth it adding a failure interceptor to the base?

mcousillas6: I just didn't use it on the base

src/App.js
```diff
@@ -1,12 +1,16 @@
...
 import { sessionService } from 'redux-react-native-session';
+import api from 'api';
+import applyDefaultInterceptors from 'api/utils/applyDefaultInterceptors';
 
 import AppContainer from './navigators';
 import configureStore from './store/configureStore';
 
 const store = configureStore({});
 
+applyDefaultInterceptors(api);
```
agustito37: I don't quite get this default interceptors concept, why should I create Interceptors outside this module? seems to me that all the Interceptors will be defined inside this module; maybe I'm missing something.

src/api/utils/applyDefaultInterceptors.js
```diff
@@ -0,0 +1,45 @@
+import { sessionService } from 'redux-react-native-session';
+import saveSessionHeaders from './saveSessionHeaders';
+
+const ACCESS_TOKEN = 'access-token';
+
+const UNAUTHORIZED = 401;
+
+const defaultRequestInterceptors = [
```
agustito37: I don't quite get this default interceptors concept, why should I create Interceptors outside this module? seems to me that all the Interceptors will be defined inside this or imported here; maybe I'm missing something.

mcousillas6: I just put it here as a helper, since they are the interceptors that come "by default" on the base to handle session to have some way to set them up easily.

The `applyInterceptors` idea comes from inverting the dependency to keep things decoupled and avoid cases of import loops.

But you are free to call `apiService.request/responseInterceptors.use()` anywhere you want. Usually, you would want it to run before any request (and just once) to get everything set up.

src/api/utils/InterceptorManager.js
mcousillas6: Right now, on any, but it seems necessary for api completeness, otherwise you would have `.use` to add middlewares, but no way to remove them. Seemed kind of weird

src/api/utils/applyDefaultInterceptors.js
agustito37: Can you think of a case where you want to define an Interceptor after the app starts running? 
What I was thinking is that it could be that in most cases I would want to define Interceptors at the same moment you are defining the default Interceptors; so instead of applyDefaultInterceptors this could be converted to applyInitializingInterceptors or something like that.

mcousillas6: Analytics maybe? Like, only set up an interceptor if you are logged in.

agustito37: Yes, that have sense hihi.

PR #86, opened 2019-08-21
Removes redux form

mcousillas6:
Adds a basic `useForm` hook that just defines form behaviour (no ui, no validations).

- No dependencies, so it's simple to remove and integrate any other form framework if the project needs something with more features.
- Removed redux forms.
- Refactored signUp and login forms to use the hook.
- Kept `validate.js` as an example of how you could set up validations using that and the hook, but you can use whatever you want there.


src/components/SignUpForm/index.js
```diff
@@ -1,41 +1,61 @@
...
+    },
+    [onSubmit],
+  );
+
+  return (
+    <>
+      <Input
+        label={translate('SIGN_UP.email')}
+        value={values.email}
+        onChangeText={text => handleValueChange(FIELDS.email, text)}
```
fsarachu: What do you think about providing a `inputProps` function from the `useForm` hook, so on every input you have to just do something like this:
```jsx
<Input
    label={translate('SIGN_UP.email')}
    { ...inputProps(FIELDS.email) }
/>
```

Being inputProps something like this:

```js
const inputProps = (field) => ({
    value: values[field],
    onChangeText: text => handleValueChange(field, text),
});
```

fsarachu: I know this would require some workaround for other type of inputs, but I think it's doable

mcousillas6: It could be a good idea, but I don't think that belongs on the form Hook, since it would be a really specific cases. The idea of hooks is to encapsulate behaviour, helper functions should go separate. I'll add it as a separate hook

src/hooks/useForm.js
```diff
@@ -0,0 +1,44 @@
...
+        ...values,
+        [key]: value,
+      };
+      setValues(newValues);
+      runValidations(newValues);
+    },
+    [values, runValidations],
+  );
+
+  return [values, setValues, errors, setErrors, handleValueChange, handleSubmit];
```
peter7z: What about an object? I think this array is a bit overwhelming.

src/actions/userActions.js
```diff
@@ -34,18 +54,17 @@ export const logout = () => async dispatch => {
     sessionService.deleteSession();
     sessionService.deleteUser();
     dispatch(logoutSuccess());
-  } catch (err) {
-    throw err;
+  } catch (error) {
+    dispatch(logoutError(error));
```
peter7z: I prefer just the throw. I think form errors should be managed at the component level.

src/hooks/useForm.js
```diff
@@ -9,10 +9,7 @@ const useForm = (
 
   const handleSubmit = useCallback(() => {
     const newErrors = validator(values) || {};
-    const valid =
-      Object.values(newErrors)
-        .flat()
-        .filter(error => !!error).length === 0;
+    const valid = !Object.values(newErrors).flat().length;
```
fsarachu: Don't you need to still filter out falsy values before checking `.length`?

mcousillas6: nope, since the errors object omits the keys that don't have errors

src/actions/userActions.js
mcousillas6: Throwing inside an action is a really bad pattern, since you are relying on it bubbling up to your component, and then either catching it inside an errorBoundary or a try/catch block.

Moreover, this is not a form error, it's an API error from the login request, so it should be handled by redux, same way we handle successful responses.

src/hooks/useForm.js
mcousillas6: Yep, sure, I was convinced the standard was to return arrays, but seems that's not the case, so no problem

src/hooks/useValidation.js
```diff
@@ -0,0 +1,6 @@
+import { useCallback } from 'react';
+import validate from 'validate.js';
+
+validate.validators.presence.options = { allowEmpty: false };
+
+export default constraints => useCallback(values => validate(values, constraints), [constraints]);
```
aguscha333: You are not allowing for the user to pass options to `validate`  function. At the top of my head, I can think of a common option we pass: `fullMessages: false` to disable that the field name is prepended to the error message

src/store/configureStore.js
```diff
@@ -1,29 +1,39 @@
...
 import AppReducer from 'reducers';
 
 /* eslint-disable */
 const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 /* eslint-enable */
 
+const persistConfig = {
+  key: 'root',
+  storage: AsyncStorage,
+  whitelist: ['session'],
```
mariosouto: shouldn't this be persisted as well? 🤔

src/services/userService.js
```diff
@@ -1,17 +1,17 @@
...
   }
 
-  static signUp(user) {
+  signUp(user) {
     return api.post('/users', user);
   }
 }
 
-export default UserService;
+export default new UserService();
```
mariosouto: i think it was already discussed but not sure if it makes sense to use a class if it doesn't have any state. Just a suggestion though

src/hooks/useForm.js
```diff
@@ -0,0 +1,83 @@
...
+  const [values, setValues] = useState(initialValues);
+  const [errors, setErrors] = useState({});
+  const [blurredFields, setBlurredFields] = useState({});
+
+  const handleSubmit = useCallback(() => {
+    const newErrors = validator(values) || {};
+    const valid = !Object.values(newErrors)
+      .filter(error => !!error)
+      .reduce((acc, error) => {
+        error.forEach(e => acc.push(e));
```
mariosouto: why is this necessary? i'm not following it

src/hooks/useForm.js
```diff
@@ -0,0 +1,83 @@
...
+      }, []).length;
+    if (valid) {
+      onSubmit(values);
+    } else {
+      setErrors(newErrors);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [values, setErrors, validator, onSubmit, ...dependencies]);
+
+  const runValidations = useCallback(
```
mariosouto: not sure if this should be part of the hook, could be on a separate file? i'm not sure if it's aligned with the other responsibilities, wdyt?

src/store/configureStore.js
mcousillas6: This is persisted, its a whitelist

src/hooks/useForm.js
mcousillas6: This flattens the errors for each field, since you could have more than one.

src/hooks/useForm.js
mcousillas6: The runValidations (as the name suggests) is in charge of just running the validator function that is passed to the hook, so the actual validation lives outside of the hook.

src/hooks/useForm.js
```diff
@@ -0,0 +1,83 @@
+import { useState, useCallback } from 'react';
```
aguscha333: 👏

tests/screens/LoginScreen.spec.js
```diff
@@ -34,113 +27,15 @@ describe('<LoginScreen />', () => {
...
   it('should display an email input field', () => {
-    expect(emailInput.props().name).toEqual('email');
+    expect(emailInput.props().label).toEqual('Email');
   });
 
   it('should display a password input field', () => {
-    expect(passwordInput.props().name).toEqual('password');
-  });
-
-  describe('submit Login with valid email/password inputs', () => {
```
aguscha333: why are you removing all these test cases?

tests/screens/SignUpScreen.spec.js
```diff
@@ -34,141 +26,20 @@ describe('<SignUpScreen />', () => {
...
   it('should display a password input field', () => {
-    expect(passwordInput.props().name).toEqual('password');
+    expect(passwordInput.props().label).toEqual('Password');
   });
 
   it('should display a password confirmation input field', () => {
-    expect(passwordConfirmationInput.props().name).toEqual('passwordConfirmation');
-  });
-
-  describe('submit SignUp with valid email/password inputs', () => {
```
aguscha333: same

tests/screens/LoginScreen.spec.js
mcousillas6: I've removed most of the test cases because they were directly tied to redux-forms view hirearchy and method api (most checks used `submit`, `isValid`, etc.). 

Ideally, I should write new tests for these, but since  we are still evaluating FinalForm and Formik for the future I'm not sure it's worth 🤷‍♂

mcousillas6: Also, we need to review the whole test setup, because it's enzyme is still failing randomly. Maybe we could try and do something with detox, cypress style.

PR #92, opened 2019-09-03
Added android build flavors

mcousillas6:
Adds missing flavors on android to match the build targets that we have setup on iOS


agustito37:
Can you add the android commands as we have with ios:staging, develop, etc?

mcousillas6:
> Can you add the android commands to the package.json as we have with ios:staging, develop, etc?

Done!

PR #93, opened 2019-09-17
Upgraded react-navigation to v4.0.5

mcousillas6:
Migrated react-navigation to v4.0.5 

- Migration guide: https://github.com/react-navigation/react-navigation/releases/tag/v4.0.0
- Added `animatedSwitchNavigator` to have a nicer transition between navigators.
- Removed unnecesary code from build.gradle (related to react-native-navigation).
- Ugraded lodash to remove security warnings while running `npm audit`.


ios/Podfile.lock
```diff
@@ -83,10 +83,12 @@ PODS:
     - React-Core (= 0.60.5)
   - ReactNativeLocalization (2.1.5):
     - React
-  - RNCAsyncStorage (1.6.1):
+  - RNCAsyncStorage (1.5.1):
```
mato2593: Why did you downgrade this library?

mcousillas6: That's because I locked the versions, will upgrade it now, good catch

PR #97, opened 2019-10-16
Adds redux-persist and removes redux-session

mcousillas6:
Adds redux persist basic configuration for local storage.

### Changes
- Added basic whitelist configuration for the reducers that we want to persist.
- Removed react-native-redux-session since it was redundant to store things twice.
- Migrated all userActions that were using `react-native-redux-session`
- Updates api middleware to store session.

### Advantages
- Easier control for session dispatchs. With `react-native-redux-session` actions were dispatched unexpectedly in some cases.
- Ability to have cache for other reducers/entities (a feed for example).
- Migration control over persisted data.


src/actions/userActions.js
```diff
@@ -54,43 +56,37 @@ const signUpError = error => ({
...
     const { data } = await userService.login({ user });
-    await sessionService.saveUser(data.user);
-    dispatch(loginSuccess());
-  } catch (err) {
-    dispatch(loginError(err));
-    throw new SubmissionError({
-      _error: err.data.error,
-    });
+    dispatch(loginSuccess(data.user));
+  } catch (error) {
```
mariosouto: you could destructure `data` here

src/actions/userActions.js
```diff
@@ -54,43 +56,37 @@ const signUpError = error => ({
...
     dispatch(logoutRequest());
     await userService.logout();
-    sessionService.deleteSession();
-    sessionService.deleteUser();
     dispatch(logoutSuccess());
-  } catch (err) {
-    dispatch(logoutError(err));
-    throw err.data.error;
+  } catch (error) {
+    dispatch(logoutError(error.data.error));
```
mariosouto: Same

src/actions/userActions.js
```diff
@@ -54,43 +56,37 @@ const signUpError = error => ({
...
-    sessionService.saveUser(data.user);
-    dispatch(signUpSuccess());
-  } catch (err) {
-    dispatch(signUpError(err));
-    throw new SubmissionError({
-      _error: err.data.error,
-    });
+    dispatch(signUpSuccess(data.user));
+  } catch (error) {
+    dispatch(signUpError(error.data.error));
```
mariosouto: Same

src/store/configureStore.js
```diff
@@ -1,29 +1,39 @@
...
-      predicate: (getState, { type }) => !_.startsWith(type, '@@redux-form'),
     });
     middlewares.push(logger);
   }
 
+  const persistedReducer = persistReducer(persistConfig, AppReducer);
+
   const store = createStore(
-    AppReducer,
+    persistedReducer,
```
peter7z: Do we want to persist every piece of the app state by default?

src/services/userService.js
```diff
@@ -1,17 +1,17 @@
...
   }
 
-  static signUp(user) {
+  signUp(user) {
     return api.post('/users', user);
   }
 }
 
-export default UserService;
+export default new UserService();
```
peter7z: Why this change? This class holds no state.

src/screens/MainScreen/index.js
```diff
@@ -21,7 +21,7 @@ const MainScreen = ({ navigation }) => {
 
   return (
     <View style={styles.container}>
-      <Text>Hey {email}, you&#39;re logged in!</Text>
+      <Text>Hey {email || ''}, you&#39;re logged in!</Text>
```
peter7z:
```suggestion
      <Text>Hey{` ${email}` || ''}, you&#39;re logged in!</Text>
```

src/store/configureStore.js
mcousillas6: Nope, we have a whitelist in `cofingureStore.js` that specifies which reducers are going to be persisted, by default I just left `session`

src/services/userService.js
mcousillas6: For consistency sake. The `ApiService` exports a singleton, so it made more sense to keep all services exposing the same interface.

src/reducers/sessionReducer.js
```diff
@@ -0,0 +1,27 @@
...
+  user: null,
+  info: null,
+};
+
+const handleLoginSuccess = (state, { user }) => {
+  state.user = user;
+};
+
+const handleLogoutSuccess = state => {
+  state.user = null;
```
aguscha333: I think here you can just do `return initialState;`

mcousillas6: I think I can't since this is an immer callback and we need to mutate the state without returning it.

aguscha333: I am using immer and have this on logout success:
image

mcousillas6: neat, will change it then

PR #111, opened 2020-02-07
Added react-navigation v5

acollazomayer:
Update react navigation to version 5
Bumped react native version to 0.61.5
Bumped some library versions


src/navigators/AppStack.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { createStackNavigator } from '@react-navigation/stack';
+
+import { MAIN_SCREEN } from 'constants/screens';
+
+import MainScreen from 'screens/MainScreen';
+
+const Stack = createStackNavigator();
```
kevquincke: Can you deconstruct here?
```suggestion
const { Navigator, Screen } = createStackNavigator();
```

src/navigators/AuthStack.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { createStackNavigator } from '@react-navigation/stack';
+
+import { LOGIN_SCREEN, SIGN_UP_SCREEN } from 'constants/screens';
+
+import LoginScreen from 'screens/LoginScreen';
+import SignUpScreen from 'screens/SignUpScreen';
+
+const Stack = createStackNavigator();
```
kevquincke: Same here

src/navigators/index.js
```diff
@@ -1,29 +1,28 @@
...
-  {
-    AppLoader,
-    AuthStack,
-    AppStack,
-  },
-  {
-    initialRouteName: 'AppLoader',
-  },
-);
+const Stack = createStackNavigator();
```
kevquincke: And here

src/navigators/index.js
```diff
@@ -1,29 +1,28 @@
...
 
-export default createAppContainer(AppNavigator);
+const Navigation = () => {
+  const { user, info } = useSession();
+
+  return (
+    <NavigationContainer>
+      <Stack.Navigator headerMode="none">
+        {user && info ? (
+          <Stack.Screen name="AppStack" component={AppStack} />
```
mcousillas6: Just a suggestion, but it's a good idea to have all screen/stacks names on constants, so it's easier to match names and navigations

acollazomayer: Cool

src/navigators/AppStack.js
acollazomayer: I am not so sure to do this. It might happen that we want to use a Tab navigation inside of a stack navigation and the Internal component names will Collide. Tab navigation also has Screen and Navigator. I think this verbosity is beneficial.

mcousillas6: Agree with maco here, I think it's a similar case to `Animated.View` where you want to have the prefix to avoid collisions

PR #114, opened 2020-03-02
Feature/added native testing library

acollazomayer:
- Added Native testing library [](https://www.native-testing-library.com/)
- Fix some bugs with network errors making the app crash.
- Removed Enzyme and JSDOM


src/components/LoginForm/index.js
```diff
@@ -38,12 +38,19 @@ const LoginForm = ({ onSubmit }) => {
         label={strings.SIGN_IN.email}
         keyboardType="email-address"
         autoCapitalize="none"
+        testID="email-input"
```
kevquincke: Will this also work on Android or should we make a wrapper?

acollazomayer: The wrapper is only needed if we want to do e2e. By the time we reach to that point the bug with android will probably be fixed.

tests/components/LoginForm.spec.js
```diff
@@ -0,0 +1,131 @@
...
+
+    describe('when the email input is not valid', () => {
+      beforeEach(() => {
+        fireEvent.changeText(input, 'example');
+        fireEvent.blur(input);
+      });
+
+      it('should show a email is not valid error', () => {
+        expect(wrapper.queryAllByLabelText('form-error')).toHaveLength(1);
+        expect(wrapper.queryByText('Email is not a valid email')).toBeTruthy();
```
aguscha333: what do you think about defining the errors as an object or constants at the top of the file so you are not repeating the hardcoded versions throughout the file?

tests/components/LoginForm.spec.js
```diff
@@ -0,0 +1,131 @@
...
+      beforeEach(() => {
+        fireEvent.blur(input);
+      });
+
+      it('should show a required error', () => {
+        expect(wrapper.queryAllByLabelText('form-error')).toHaveLength(1);
+        expect(wrapper.queryByText("Email can't be blank")).toBeTruthy();
+      });
+    });
+  });
```
aguscha333: I would group the cases by ordering them so that all cases that check that an error appears for an input are together. Makes it easier to follow.

tests/components/LoginForm.spec.js
```diff
@@ -0,0 +1,131 @@
...
+      beforeEach(() => {
+        fireEvent.blur(input);
+      });
+
+      it('should show a required error', () => {
+        expect(wrapper.queryAllByLabelText('form-error')).toHaveLength(1);
+        expect(wrapper.queryByText("Password can't be blank")).toBeTruthy();
+      });
+    });
+  });
```
aguscha333: Same, try to keep consistency with the order of cases.

tests/components/SignUpForm.spec.js
```diff
@@ -0,0 +1,186 @@
...
+      beforeEach(() => {
+        fireEvent.blur(input);
+      });
+
+      it('should show a required error', () => {
+        expect(wrapper.queryAllByLabelText('form-error')).toHaveLength(1);
+        expect(wrapper.queryByText("Email can't be blank")).toBeTruthy();
+      });
+    });
+  });
```
aguscha333: Is there a way to reuse this? it's the exact same code as in the LoginForm Spec

tests/navigators/Navigator.spec.js
```diff
@@ -0,0 +1,125 @@
...
+            email: 'example@rootstrap.com',
+          },
+          info: {
+            token: 'token',
+            uid: 'example@rootstrap.com',
+            client: 'client',
+          },
+        },
+      });
+      wrapper = renderWithRedux(<Navigator />, store);
```
aguscha333: this looks like something that would be useful as a helper since most of the screens in real projects will require the user to be already logged in

tests/navigators/Navigator.spec.js
```diff
@@ -0,0 +1,125 @@
...
+        mockedHttpClient(store)
+          .onDelete('/users/sign_out')
+          .reply(
+            200,
+            { success: true },
+            {
+              'access-token': 'token',
+              uid: 'example@rootstrap.com',
+              client: 'client',
+            },
```
aguscha333: You are using this a lot, could be a constant

tests/components/SignUpForm.spec.js
acollazomayer: I would prefer not to this. I've been reading about testing coding style and some authors agreed that is better to be more explicit on tests, if you have duplicate code that tests similar functionality is better to be more verbose about it. Extracting the logic can lead to false positives and can be more difficult to track.

tests/screens/SignUpScreen.spec.js
```diff
@@ -1,45 +1,106 @@
...
+            },
+          );
+        fireEvent.press(wrapper.queryByTestId('signup-submit-button'));
+
+        await wait(() => expect(wrapper.queryAllByLabelText('form-error')).toEqual([]));
+      });
+    });
+
+    describe('if there is a network error', () => {
+      it('should show no errors', async () => {
```
peter7z: For me `something went wrong` is an error...

tests/screens/MainScreen.spec.js
```diff
@@ -0,0 +1,32 @@
...
+      session: {
+        user: {
+          email: 'example@rootstrap.com',
+        },
+      },
+    });
+    wrapper = renderWithNavigation(MainScreen, store);
+  });
+
+  it('should render the login screen', () => {
```
mato2593: Shouldn't it be `the main screen`?

tests/navigators/Navigator.spec.js
```diff
@@ -0,0 +1,125 @@
+import React from 'react';
+import { fireEvent, wait } from '@testing-library/react-native';
+
+import { LOGIN_SCREEN, MAIN_SCREEN } from 'constants/screens';
+import Navigator from 'navigators';
+
+import { renderWithRedux, mockedHttpClient, configureStore } from '../helpers';
+
+describe('Navigator', () => {
+  describe('when the is logged out', () => {
```
mato2593: `user` missing

tests/navigators/Navigator.spec.js
mato2593: You should remove the headers since you are logging the user out

tests/screens/LoginScreen.spec.js
```diff
@@ -1,41 +1,99 @@
...
+
+        await wait(() => {
+          expect(wrapper.queryAllByLabelText('form-error')).toHaveLength(1);
+          expect(wrapper.queryByText('Invalid login credentials. Please try again.')).toBeTruthy();
+        });
+      });
+    });
+
+    describe('if there is a network error', () => {
+      it('should show no errors', async () => {
```
mato2593: `should show errors`

PR #116, opened 2020-04-03
iOS splash screen

elizabethlofredo:
### Description
iOS requires for the Splash screen to be `.storyboard` instead of `.xib` as we currently have. This is going to be required to submit the app for review since June 30th.  You can check the [annoucment]( https://developer.apple.com/news/?id=01132020b).


tamarasuarezrod:
Did you add it in both targets? (develop and staging)

elizabethlofredo:
> Did you add it in both targets? (develop and staging)

@tamarasuarezrod yes!

PR #118, opened 2020-06-08
enhancement/ useForm

elizabethlofredo:
Due to changes in another project, I noticed that the `validateOnChange` function from the useForm hook was triggering the errors when you start typing. 

* These changes include an array with the `touched` fields to trigger the errors from the validateOnChange after the first onBlur.
* I change the Login and SignUp forms to use this type of validation. 

Comments are welcome! 🙂 and if this is ok with everyone I will make this small enhancement on the react-base too


elizabethlofredo:
I will fix the tests 😅

PR #123, opened 2020-07-28
Upchrore: date circleCi config to use orbs + lock react-navigation version

aguscha333:
This PR updates the circleCi configuration to use [orbs](https://circleci.com/orbs/)

It also locks the `react-navigation` versions to use `5.0.0` instead of being `^5.0.0`. This makes it much more stable and will save us headaches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,26 +1,23 @@
-version: 2
+version: 2.1
+orbs:
+  node: circleci/node@1.1.6
 jobs:
-  build:
+  build-lint-and-test:
     environment:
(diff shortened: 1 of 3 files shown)


aguscha333:
Sorry for the double PR, had created it from my fork and circleCi was not running

aguscha333:
@peter7z Tested and `react-navigation` `5.7.0` breaks the tests as well 😞 . I think I encountered this error actually because I had to switch to npm and it decided differently which version of the library to use since there was no `package-lock.json` to begin with.

I wanted to upload that fix so nobody else wastes time like me not knowing what is going on. But yeah, we should for sure try to upgrade and fix the tests on another PR.

PR #124, opened 2020-08-11
Updated iOS Fastlane and added Android Fastlane config

sebalopez:
* Added lanes for iOS AdHoc and Appstore QA builds 
  - expects Certificates and profiles available locally 
  - requires Apple authentication present in env vars
* Added lanes for Android build and publish 
  - expects Keystore available locally
* Included steps for pushing Dev/QA builds into S3 
  - requires AWS credentials and target present in env vars
* Included steps for sending Slack notifications
  - required Slack webhook and channel defined in env vars
* Included detailed usage information in READMEs


ios/fastlane/Fastfile
```diff
@@ -15,164 +15,206 @@
...
     )
   end
 
-  lane :build_staging do
-    sync_code_signing(
-      git_url: certificates_git_url,
-      username: username,
-      type: 'appstore'
+  # Generate Appstore build and push to TestFlight - modify scheme as appropiate
+  lane :release_qa_appstore do
```
aguscha333, line 178: same comment I had on the android file. I think that it is worth having a separate build target for QA if that is sth that we are aiming to have for release in the base.

Dev build target and release can also be useful to set up since some times there are some issues that only happen when testing via testflight or google play console and it is best to have an app that you can update and not break what QA or staging are testing.

android/fastlane/Fastfile
```diff
@@ -0,0 +1,180 @@
...
+default_platform(:android)
+
+lane :build do |options|
+  # increment version
+  version = get_version_name(
+      gradle_file_path: './app/build.gradle'
+  ).to_s
+  build_number = increment_version_code(
+      gradle_file_path: './app/build.gradle'
+  ).to_s
```
aguscha333, line 40: if I release a version to QA and then subsequently release the same version to staging, will they have different build_numbers? Is that the intended behaviour?

android/fastlane/Fastfile
```diff
@@ -0,0 +1,180 @@
...
+#
+# For a list of all available plugins, check out
+#
+#     https://docs.fastlane.tools/plugins/available-plugins
+#
+
+# Uncomment the line if you want fastlane to automatically update itself
+# update_fastlane
+
+default_platform(:ios)
```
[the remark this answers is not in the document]
sebalopez: *facepalm* you're right, sorry! great catch :)

android/fastlane/Fastfile
```diff
@@ -0,0 +1,180 @@
...
+    slack_url: slack_url,
+    use_webhook_configured_username_and_icon: true,
+    fail_on_error: false,
+    success: true
+  )
+end
+
+lane :release_qa_appstore do
+  
+  build(flavor: 'Dev')
```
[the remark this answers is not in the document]
sebalopez: good point, I have seen projects use Dev flavor for QA but it makes sense to assume that projects are following the right branch strategy

ios/fastlane/Fastfile
```diff
@@ -15,164 +15,206 @@
...
-      export_method: 'app-store'
+  lane :release_staging do
+    ensure_git_status_clean
+    build_and_sign(
+      type: 'appstore',
+      scheme: app_name+'-Staging',
+      method: 'app-store'
+    )
+    publish(
+      scheme: app_name+'-Develop'
```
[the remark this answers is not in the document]
sebalopez: *facepalm2* good catch again!

android/fastlane/Fastfile
sebalopez: from what I've seen so far build numbering across targets do not keep much relation to each other. But in any case I am thinking about removing the increment_version_code altogether and leave this for manual update as we are not commiting this (won't play well with CICD) and it doesn't really add value otherwise

ios/fastlane/Fastfile
sebalopez: gotcha. Will add Dev releases and restructure the fastfiles a bit to minimize duplicated code

ios/fastlane/Fastfile
```diff
@@ -63,24 +64,64 @@ platform :ios do
...
       output_name: options[:scheme]+".ipa",
       export_team_id: team_id
     )
   end
 
-  lane :publish do |options|
+  # Generate AdHoc build and push to S3
+  lane :publish_adhoc do |options|
+    version = get_info_plist_value(
+      path: app_name+"/Info.plist",
```
aguscha333, line 88: As I have it on my project, each build target has it's own info.plist. The only difference that I see is that they have a different `Bundle display name`. I think that the actual name that is displayed for the app is `Product Name` on the Build Settings.
Can you make sure that the `Bundle display name` in the Info.plist has `$(PRODUCT_NAME)` as value and that the correct name is displayed for the envs? testing it with one env that is not production.

agustito37 (approved):
Seems cool!

elizabethlofredo (approved):
Nothing to add besides Agus's comments 👌

PR #126, opened 2020-09-11
Replace CircleCI with GitHub Actions

sebalopez:
Replace the steps in .circle/config.yml with .github/workflows/test.yml:
* install dependencies
* run unit tests
* run linter
* send to CodeClimate

Upon merge of this PR, webhook for Circle needs to be removed


.github/workflows/test.yml
```diff
@@ -0,0 +1,54 @@
+name: CI Test 
+
+# Run for any commits to any branch
+on: [push, pull_request]
```
aguscha333, line 4: I'm not quite sure if with circleCi the tests where run on every branch on push or just PRs, perhaps they were but we just weren't seeing the results cause they weren't attached to a PR 🤔 . If we ever want to minimize the number of minutes we use for the projects we can always restrict it to push to main branches and pull_request: types: [opened, synchronize]

aguscha333 (approved):

.github/workflows/test.yml
aguscha333: I just saw that in the checks of the PR you have the same check twice, If we are checking on push, then it also covers when a pr is created and `pull_request` should not be needed.

sebalopez: The workflow runs upon the push to the branch (on the branch code) and upon creation of the PR (on the code resulting from the merge of the branch into master). The PR checks are actually set like this automatically by GH. 
We can specify which -if any- should be required to pass in order to approve the PR.

aguscha333 (approved):

PR #130, opened 2020-09-22
feat: upgrade react native to 0.63.2 and fix android release

aguscha333:
### Description
This PR upgrades the current version we are using of React Native 0.62.2 to the newest version 0.63.2.

This is upgrade is mainly being done because:
- we currently have for android the `targetSdkVersion` at 28 and in order to release apps through the google play console we need to have it at 29. This newest version of react native uses that `targetSdkVersion` and it's good to follow react native upgrade diff for the upgrade just in case there are any other changes needed.

- More importantly, we currently have an issue with the Flipper version that we are using. It basically breakes the APK generation because there is a file mising for release mode. Flipper version is a bit harder to upgrade without upgrading react native since it touches a lot of configuration files.

I tested debug and release for both platforms.


ManuViola77 (approved):
🧞‍♀️ 👏 💜

missx (approved):
👏

tamarasuarezrod (approved):

PR #131, opened 2020-09-23
Enhancement/upgrade use form hook

aguscha333:
### Description

This PR contains a couple of upgrades to the usage of forms. Here is a list:

* Add active state support for inputs, I changed the input border color when active to display this
* Add the possibility to disable the submit button if the form is not valid, which required
	* Adding a useEffect hook in the input to register it in the values of useForm
	* Adding the touched flag so we can show errors only if the user has interacted with the field
	* **UPDATE:** on submit touch all keys with errors just in case the submit button is not disabled if the form is not valid
* Add a flag that lets you validate all the fields when one changes, useful if fields depend on others
* Swapped validate.js for [Rootstrap's own validation lib](https://github.com/rootstrap/validate/) which is a wrapper of validate.js. It's a hub for cannon validate.js validators and also other custom validators that we have used that are so common that were calling to be grouped.

### Preview
Here is a preview of how the form now looks like with active, touched, and disabled states.

useFormActiveAndTouched

### DIFF
--- a/src/components/common/Input/styles.js
+++ b/src/components/common/Input/styles.js
@@ -3,6 +3,11 @@ import { StyleSheet } from 'react-native';
 const styles = StyleSheet.create({
   input: {
     height: 50,
+    borderColor: '#CCCCCC',
+    borderWidth: 1,
+  },
+  inputActive: {
+    borderColor: '#666666',
   },
   button: {
     alignSelf: 'center',
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@react-navigation/native": "5.0.0",
     "@react-navigation/stack": "5.0.0",
     "@rootstrap/redux-tools": "^0.4.2",
+    "@rootstrap/validate": "^1.0.0",
     "axios": "^0.19.2",
     "eslint-plugin-prettier": "3.1.0",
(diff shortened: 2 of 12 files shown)


ManuViola77 (approved):
👏

src/components/common/Input/styles.js
```diff
@@ -3,6 +3,11 @@ import { StyleSheet } from 'react-native';
 const styles = StyleSheet.create({
   input: {
     height: 50,
+    borderColor: '#CCCCCC',
+    borderWidth: 1,
+  },
+  inputActive: {
+    borderColor: '#666666',
```
missx, line 10: Can we store these as constants?

missx (approved):
👏

src/components/common/Input/styles.js
aguscha333: I thought about it and I decided against it. This file already has a hardcoded color and I think that the reasoning behind it is that these styles are just placeholders and are very likely to change so just having them here makes it easier for the dev to remove and implement the real styling. But I am open to suggestions.

missx: Got it, seems a good call

PR #132, opened 2020-09-25
Include Sonarqube integration

sebalopez:
Include Sonar static analysis and test coverage reporting in CI build


.github/workflows/test.yml
```diff
@@ -5,6 +5,9 @@ on: [push, pull_request]
 
 env:
   CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+  SONAR_PROJECT: react-native-base
```
lucasocon, line 8: Should we need the SONAR_PROJECT key here?

sebalopez: I guess it is not strictly necessary, we could just reference the project name directly in the sonar execution, it is just cleaner to modify later

pablanco: @sebalopez you could add this as another secret VAR, so to replay this example we will just need to change secrets, instead of secrets and config code.

pablanco (approved):

PR #133, opened 2020-11-23
fix: remove unnecessary dependencies

aguscha333:
This PR removes 2 dependecies:
- `redux-form` since we are not using it anymore, we have our own form hooks.

- `reselect` since we are not really using it on projects either. Now with `react-redux`'s `useSelector` and being able to create our own hooks in case we are selecting something from the store often, I don't see much use for having `reselect` on the base, and I don't see use of it on recent Rootstrap projects.


manema (approved):
👌

agustito37 (approved):

missx (approved):

nsantos16 (approved):

jpirazusta (approved):

elizabethlofredo (approved):

PR #134, opened 2020-11-23
fix: update sonar scanner version

aguscha333:
Fix failing test action as mentioned by @pablanco that we just need to upgrade de version being used


elizabethlofredo (approved):

pablanco (approved):
Nice!!

PR #142, opened 2021-02-18
feat: add version bump script

aguscha333:
## Description
When sending a new version there is a manual step that we can never skip, that is bumping the version number of our app. Right now that means going to 4 different places if you count `dev`, `staging`, and `prod` for Android and iOS.

To help ensure that the versions for all envs are updated and to make it easier on the devs as well, I created a a script that can update the version for you in all places needed. Just select the type of version bump you want and done.

This script is heavily inspired in the lib [rnbv](https://www.npmjs.com/package/rnbv) which unfortunately does not support build targets for iOS and it also doesn't bump the android version code so we couldn't just use that.

For usage instructions see the changes in the readme file.

If you are working on an existing project and want to use this script, it should be fairly simple to apply the changes in this PR to it. Just make sure to update the `rnbv.config.js` file with the correct paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,10 @@
 You can also create `.env.prod` and `.env.staging` to define environment variables for production and staging.
 
 3. Rename your new project using [react-native-rename](https://github.com/junedomingo/react-native-rename). You will also need to rename other files manually.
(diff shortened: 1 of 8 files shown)


ManuViola77 (approved):

manema (approved):
🚀🚀

elizabethlofredo (approved):
👏

jdalto (approved):
🚀 🚀 👏 👏

PR #157, opened 2021-08-03
fix locale

Stukz:
### Description:
The current locale implementation have issues with Going To/Find definitions, it was related that the alias used also was referenced from react-native core types definition.

In order to make it work a rename to the folder & alias was necessary.

---

#### Tasks:

- [x] Tested on iOS
- [ ] Tested on Android
- [ ] Tested accessibility
- [ ] Added tests

### Preview:
As you can see, when I use my shortcut to go to definition it shows the preview & I can jump right on to the file.

https://user-images.githubusercontent.com/57040392/128077074-2e2a7b3b-6439-475e-9800-80b3dcb0ee32.mov


ManuViola77 (approved):
Awesome! 👏🏻  🧞‍♂️

jpirazusta (approved):
👏

PR #170, opened 2022-04-01
chore: exclude environment, coverage & package-lock files

Stukz:


aguscha333 (approved):

jpirazusta (approved):

.gitignore
```diff
@@ -35,6 +35,7 @@ local.properties
 node_modules/
 npm-debug.log
 yarn-error.log
+package-lock.json
```
asdolo, line 38: Hmm. I know the readme clearly says to use `yarn`, but it doesn't mention you *can't* use `npm`. So if anyone uses `npm` and they don't pay any attention to the `package-lock.json` file being ignored it could be a problem.

@Stukz @aguscha333 @jpirazusta @fernandatoledo @juanchoperezj what do you think?

asdolo (changes requested):

.gitignore
juanchoperezj: I agree too, we can keep tracking `package-lock.json`

Stukz: fixed!

juanchoperezj (approved):

PR #171, opened 2022-04-01
chore: install react navigation and imports sort plugin

Stukz:
### Description
This PR includes:

1. installation of React Navigation and initial setup
2. installation of  [prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports) to define a standard order for the imports

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,7 +1,12 @@
 module.exports = {
   arrowParens: 'avoid',
   bracketSameLine: true,
-  bracketSpacing: false,
+  bracketSpacing: true,
   singleQuote: true,
   trailingComma: 'all',
+  printWidth: 100,
+  jsxBracketSameLine: true,
+  importOrder: ['^@(.*)/(.*)$', '^screens/(.*)$', '^constants/(.*)$', '^[./*]'],
+  importOrderSeparation: true,
+  importOrderSortSpecifiers: true,
 };
--- a/src/screens/WelcomeScreen/index.tsx
+++ b/src/screens/WelcomeScreen/index.tsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import { SafeAreaView, ScrollView, StatusBar, Text, View, useColorScheme } from 'react-native';
+import {
+  Colors,
+  DebugInstructions,
+  Header,
+  LearnMoreLinks,
+  ReloadInstructions,
+} from 'react-native/Libraries/NewAppScreen';
+
+import styles from './styles';
+
+const Section: React.FC<{
+  title: string;
+}> = ({ children, title }) => {
+  const isDarkMode = useColorScheme() === 'dark';
+  return (
+    <View style={styles.sectionContainer}>
+      <Text
+        style={[
+          styles.sectionTitle,
+          {
+            color: isDarkMode ? Colors.white : Colors.black,
+          },
+        ]}>
+        {title}
+      </Text>
+      <Text
+        style={[
+          styles.sectionDescription,
+          {
+            color: isDarkMode ? Colors.light : Colors.dark,
+          },
+        ]}>
+        {children}
+      </Text>
+    </View>
+  );
+};
+
+const WelcomeScreen = () => {
+  const isDarkMode = useColorScheme() === 'dark';
+
+  const backgroundStyle = {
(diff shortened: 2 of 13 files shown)


.prettierrc.js
```diff
@@ -1,7 +1,12 @@
...
   bracketSameLine: true,
-  bracketSpacing: false,
+  bracketSpacing: true,
   singleQuote: true,
   trailingComma: 'all',
+  printWidth: 100,
+  jsxBracketSameLine: true,
+  importOrder: ['^@(.*)/(.*)$', '^screens/(.*)$', '^constants/(.*)$', '^[./*]'],
+  importOrderSeparation: true,
+  importOrderSortSpecifiers: true,
```
aguscha333, line 11: Nice!

src/screens/WelcomeScreen/index.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
aguscha333, line 1: I like more putting an index file with
```
export { default } from './HomeScreen'
```
and then having `HomeScreen.tsx` and `HomeScreen.styles.ts`

So there is not a hundred files named the same, it makes files easier to find. WDYT?

src/screens/WelcomeScreen/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  DebugInstructions,
+  Header,
+  LearnMoreLinks,
+  ReloadInstructions,
+} from 'react-native/Libraries/NewAppScreen';
+
+import styles from './styles';
+
+const Section: React.FC<{
+  title: string;
```
asdolo, line 14: What do you thing about separating the interface for the props?

```ts
interface SectionProps {
  title: string;
}; 

const Section: React.FC<SectionProps> = ({ children, title }) => {

  ...

}
```

src/screens/WelcomeScreen/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  Colors,
+  DebugInstructions,
+  Header,
+  LearnMoreLinks,
+  ReloadInstructions,
+} from 'react-native/Libraries/NewAppScreen';
+
+import styles from './styles';
+
+const Section: React.FC<{
```
asdolo, line 13: What's your viewpoint on declaring more than one React Component per file?

- Is it worth to move this to a separate file?
- Is it worth to do it now? Or maybe only when it's needed?

asdolo (approved):

src/screens/WelcomeScreen/index.tsx
aguscha333: yes and also it is easier to find component files if they are not all named index

asdolo: Right. On the other side we'll have to create one more file for every new component that is redundant 🤔.

src/screens/WelcomeScreen/index.tsx
asdolo: Yes, I was thinking of this as a good practice for all of us to have, and also for new people reading TypeScript code.

src/screens/WelcomeScreen/index.tsx
aguscha333: Yup, I agree that neither solution is perfect but having worked with both I prefer much more the one that I am proposing, I think it also scales better in the developer experience aspect.
We can leave it up to the developer as well, if the developer is opinionated enough on the matter they will have no problem changing it. If they aren't they will use what's in the base.

src/screens/WelcomeScreen/index.tsx
aguscha333: yuppp totally agree on that, just not sure if the change in worth in this instance

Stukz: yes It was the boilerplate from RN CLI, just to have something to render

src/screens/WelcomeScreen/index.tsx
Stukz: maybe we could bring this discussion on the slack channel and use a poll

aguscha333 (approved):

jpirazusta (approved):

PR #178, opened 2022-07-21
chore: add axios client + error handling + interceptors

Stukz:


src/network/client/error-handler.ts
```diff
@@ -0,0 +1,42 @@
+const DEFAULT_ERROR = 'something went wrong';
+
+// TODO: fix types definition based on the error response from the backend
```
Stukz, line 3: FYI: this whole file it's from the old base, keep in mind this could change on every backend project

tamarasuarezrod (approved):
👏

aguscha333 (approved):

juanchoperezj (approved):

asdolo (approved):

PR #179, opened 2022-07-21
chore: add local storage support

Stukz:


aguscha333 (approved):

src/storage/index.ts
```diff
@@ -0,0 +1,39 @@
...
+type Value = string;
+
+export enum StoreKeys {
+  session = 'session',
+}
+
+export * from './parser';
+
+class Storage {
+  #Store = new MMKV();
```
juanchoperezj, line 12: I'm curious about this, do you have a documentation/link? To look at it.

juanchoperezj (approved):

tamarasuarezrod (approved):

src/storage/index.ts
Stukz: yeah sure, @juanchoperezj @tamarasuarezrod 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

PR #180, opened 2022-08-19
Add react-query

aguscha333:
This PR adds the basic react query integration and some example models, services and queries that devs can base their queries on. The example is not functional and because of that was not tested but it was generated based on similar code from an existing project.

### DIFF
--- a/src/network/models/example-models.ts
+++ b/src/network/models/example-models.ts
@@ -1,0 +1,22 @@
+//
+//
+//
+//
+// THIS IS NOT FUNCTIONAL, JUST FOR EXAMPLE PURPOSES, REMOVE WHEN NOT NEEDED
+//
+//
+//
+export type ExampleRequest = {
+  pet: {
+    name: string;
+    age?: number;
+  };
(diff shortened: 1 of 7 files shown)


santiagorobra (approved):
🚀

src/network/models/example-models.ts
```diff
@@ -0,0 +1,22 @@
...
+  };
+};
+
+export type ExampleResponse = {
+  pet: {
+    id: string;
+    name: string;
+    age?: number;
+  };
+};
```
juanchoperezj, line 22: I saw a few days ago this library [zod](https://zod.dev/?id=basic-usage). Will depend on the use case but could be useful.

juanchoperezj (approved):
💯

src/network/models/example-models.ts
aguscha333: that is for validations, not sure if it helps you define ts models for the endpoints

marcos09 (approved):

PR #181, opened 2022-09-06
add react-native-accessibility-engine

juanchoperezj:
Screen Shot 2022-09-05 at 10 11 22 PM


__tests__/App.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+describe('App', () => {
+  it('renders correctly', () => {
+    const app = render(<App />);
+    expect(app).toBeTruthy();
+  });
+
+  it('elements should be accessible', () => {
+    const { getByText, getByA11yState } = render(<App />);
+    const text = getByText('Welcome Screen');
+    const button = getByA11yState({ disabled: false });
```
aguscha333, line 17: Getting sth by it's state seems weird 🤔. Does this engine has a `getByRole` or something similar?

aguscha333 (commented):
Nice work!

__tests__/App.test.tsx
juanchoperezj: I did this just for demonstration purposes. I'll change it to use `getByRole`.

Stukz (approved):

aguscha333 (approved):

PR #183, opened 2022-09-29
Add Prettier and ESLint settings

marcos09:
This PR adds prettier, ESLint, and accessibility  rules based on previous rules we had in our previous React Native Base


Stukz (approved):

.eslintrc.json
```diff
@@ -0,0 +1,162 @@
...
+    "react-native-a11y/has-accessibility-props": 1,
+    "react-native-a11y/has-valid-accessibility-actions": 1,
+    "react-native-a11y/has-valid-accessibility-role": 1,
+    "react-native-a11y/has-valid-accessibility-state": 1,
+    "react-native-a11y/has-valid-accessibility-states": 1,
+    "react-native-a11y/has-valid-accessibility-component-type": 1,
+    "react-native-a11y/has-valid-accessibility-traits": 1,
+    "react-native-a11y/has-valid-accessibility-value": 1,
+    "react-native-a11y/no-nested-touchables": 1,
+    "react-native-a11y/has-valid-accessibility-descriptors": 1,
```
aguscha333, line 44: where was this taken from? doesn't it have like a recommended set of rules by default?

marcos09: These rules came from [react-native-a11y](https://www.npmjs.com/package/eslint-plugin-react-native-a11y) @aguscha333 but I wasn't sure about default values for those rules so I preferred to specify them.

aguscha333: cool

PR #184, opened 2022-10-12
Add new rename command for New React Native Base using different schemes

marcos09:
## Rename App using different schemes instead of targets

- This PR is based on  [Ignite](https://github.com/infinitered/ignite) and [Ignite-cli](https://www.npmjs.com/package/ignite-cli)
- It modifies the rename command to support multiple schemes and keeps the environment name after renaming the app. 
- It allows to change the package name and build identifier when users run `yarn rename`
- When users change the package name it rename and move all native folders based on the new identifier for the app.

All rename code is inside `rnb-cli ` folder. There are a lot of files to isolate the rename from Ignite to our base using `gluegun` but the most important file is  `rename.ts`. All other files are created by `gluegun` to create new commands 

## Command use:

Screenshot 2022-10-12 at 11 05 45

## Files changes after rename the app and select a new Package name / Bundle Identifier:

Screenshot 2022-10-12 at 11 06 27


rnb-cli/src/commands/help.ts
```diff
@@ -0,0 +1,22 @@
...
+    const { meta } = toolbox
+
+    p()
+
+    heading(`Welcome to React Native Base CLI ${meta.version()}!`)
+    p()
+    heading("Commands")
+    p()
+    command("rename      ", "Renames your React Native project (experimental)", [
+      "ignite rename NewName com.mycompany.newname",
```
aguscha333, line 18: Maybe change this to `rnb-cli` instead of `ignite` we don't need to be referencing ignite here

aguscha333 (commented):
Great job!!!

Stukz (approved):

juanchoperezj (approved):
Besides Agus's comments, looks good!

aguscha333 (approved):

PR #190, opened 2022-10-20
chore: add script for bumping version

ManuViola77:
[Backlog Card](https://www.notion.so/rootstrap/Version-Bump-Standard-41dae38cd98a4aa89db422b17490209a)

This PR adds the script for bumping the version. It is based on [this library](https://github.com/llotheo/react-native-bump-version) but I modify to have an option with only bumping the build number, leaving the version number intact. 

Usage: run `yarn bump` and chose an option:
image

In that example:

- choosing Major would do this: 

image

- choosing Build Number would do this: 

image

Note that Android Code Number always goes up while in iOS the Build Number resets to 0 with all options except Build Number that will only sum 1 to the current value.

PS: I wanted to add something to the README but I see the README PR is still in code review so I will add the usage in the README file in a later PR.

PS2: In `ios/ReactNativeBase/Info.plist` file I only changed the number from 1.0 to 1.0.0, but it indented everything different, and the changes shown look worse than what it really is.


package.json
```diff
@@ -33,21 +33,26 @@
...
   },
   "dependencies": {
     "@react-navigation/native": "^6.0.8",
     "@react-navigation/native-stack": "^6.5.2",
-    "react-native-config": "^1.4.6",
     "@tanstack/react-query": "^4.2.1",
     "axios": "^0.27.2",
+    "chalk": "4.1.2",
     "humps": "^2.0.1",
+    "husky": "^8.0.1",
```
marcos09, line 46: Why did you add husky as a dependency? We agreed with the team to use Left hook instead. [Here](https://www.notion.so/rootstrap/Pre-Commit-179e78ff2a7f4c568489bc683493b155) you have the explanation

package.json
```diff
@@ -33,21 +33,26 @@
...
+    "prompts": "^2.4.2",
```
marcos09, line 47: I think prompts, chalk & semver should be a dev dependencies. WDYT?

rnbv.js
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const chalk = require('chalk');
+const prompts = require('prompts');
+const semver = require('semver');
+
+const configMap = new Map([
+  ['androidPath', './android/app/build.gradle'],
+  ['iosPath', './ios/ReactNativeBase/Info.plist'],
```
marcos09, line 10: The iOS directory could change when the app is renamed. After renaming the app it changes the location for every native iOS file to use the new app name as the new project folder.

marcos09 (changes requested):
Great Job Manu 🚀  I requested for a small changes, LMN if you have questions

rnbv.js
ManuViola77: I added the `rnbv` file to the `filesToPatch` so the renames changes the value

marcos09 (approved):
Awesome work Manu. Keep doing it

rnbv.js
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env node
```
aguscha333, line 1: Given that we now have more scripts that live at `rnb-cli/src/tools` maybe we should move this one there

ios/ReactNativeBase/Info.plist
```diff
@@ -1,55 +1,55 @@
...
-				<true/>
+				<key>localhost</key>
+				<dict>
+					<key>NSExceptionAllowsInsecureHTTPLoads</key>
+					<true/>
+				</dict>
 			</dict>
 		</dict>
+		<key>NSLocationWhenInUseUsageDescription</key>
+		<string></string>
```
Stukz, line 39: same here

Stukz (approved):

ios/ReactNativeBase/Info.plist
```diff
@@ -1,55 +1,55 @@
...
 		<dict>
-			<key>localhost</key>
+			<key>NSExceptionDomains</key>
 			<dict>
-				<key>NSExceptionAllowsInsecureHTTPLoads</key>
-				<true/>
+				<key>localhost</key>
+				<dict>
+					<key>NSExceptionAllowsInsecureHTTPLoads</key>
+					<true/>
```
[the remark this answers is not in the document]
ManuViola77: As we talked in slack, this was already here, it just changed the indentation.

ios/ReactNativeBase/Info.plist
ManuViola77: same here

aguscha333 (approved):

PR #192, opened 2022-10-26
Add/navigator types

ramirogavagnin:
Add navigation types and screens refactor


src/constants/screens.ts
```diff
@@ -1 +1,5 @@
+// Auth
 export const WELCOME_SCREEN = 'WelcomeScreen';
+
+// Main
+export const HOME_SCREEN = 'HomeScreen';
```
santiagorobra, line 5: these variables are necessary? you are already using "enum AppScreens....".

App.tsx
```diff
@@ -1,24 +1,16 @@
+import NavigationStack from 'navigation/index';
```
santiagorobra, line 1:
```suggestion
import React from 'react'
import { NavigationContainer } from '@react-navigation/native';
import { QueryClient, QueryClientProvider } from '@tanstack/react-query';

import NavigationStack from 'navigation/index';
```

We always import everything from React first, then 3rd party libraries or our own and then the other imports

App.tsx
```diff
@@ -1,24 +1,16 @@
...
 import React from 'react';
```
Stukz, line 2:
```suggestion
import React from 'react';
import NavigationStack from 'navigation/index';
```

src/navigation/stacks/auth/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { StackParamList } from 'navigation/types';
+import { AppScreens } from 'navigation/types';
+import React from 'react';
```
Stukz, line 3: if I remember correctly I had an action where the imports where automatically fixed to avoid any inconsistency

App.tsx
ramirogavagnin: totally agree, I've forgot to add the rest of path alias

src/constants/screens.ts
ramirogavagnin: totally!

santiagorobra (approved):

santiagorobra (approved):

Stukz (approved):

PR #198, opened 2022-11-08
feat: add ios app icons

ManuViola77:
This PR adds the app icons for both Android and iOS. I also reorder the names of the different envs, so the root name is the same and the env is at the end.


marcos09 (commented):
Awesome work Manu. I've never noticed how many files we need to implement different icons for each scheme. They are a lot, but it's ok, they are needed. 

Maybe I'd add a little section in the README explaining how to change the icon in Android. I assume for dev and QA environments developers will want to keep the previous ones. I think the easier way to change the icon is only replacing the png files and I think it would be helpful to say in the readme "You need to change only this" . WDYT?

Here there are [additional information](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive)

marcos09 (approved):

stivenviedman (approved):

PR #199, opened 2022-11-10
Fix eslint script

ManuelJrez:


marcos09 (approved):
Thanks for the fix. I was trying to fix it but I didn't yet and I was looking for a way to solve this. Where did you find it?

ManuViola77 (approved):
👏🏻

ManuelJrez:
@marcos09 when i realize that the problem was the rnb-cli sub-project, i search for something to avoid our linter to run in that folder (rnb-cli) and i find the .eslintignore to ignore path or files. 😅 

BTW i think we can move that `rnb-cli` maybe outside of the repo in his own repo to maintain separately and maybe adding to the base as an npm package... what do you think? 🤔

aguscha333 (approved):

aguscha333 (approved):

PR #204, opened 2022-11-24
chore: Add typing to environment variables

asdolo:
Add types for `react-native-config` module for making environment variables type-safe.

| - | Before ❌ | After ✅ |
| - | - | - |
| Type checking | image | image |
| Autocomplete/Intellisense | image | image |

### DIFF
--- a/src/@types/react-native-config.d.ts
+++ b/src/@types/react-native-config.d.ts
@@ -1,0 +1,8 @@
+declare module 'react-native-config' {
+  interface Environment {
+    API_URL: string;
+    BASE_URL: string;
+  }
+  const Config: Environment;
+  export default Config;
+}
--- a/src/network/client/index.ts
+++ b/src/network/client/index.ts
@@ -1,9 +1,5 @@
 import axios from 'axios';
-
-// TODO: replace with .env variables.
-const Config = {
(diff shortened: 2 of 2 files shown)


marcos09 (approved):

aguscha333 (approved):

juanchoperezj (dismissed):

asdolo:
@aguscha333 @marcos09 @juanchoperezj 

Sorry, I had to change [this](https://github.com/rootstrap/react-native-base/compare/03812f8e3352180a2cbb8c4df80f55e0d869a32a..a55d52e1ef76fbca7a929979ecfbcccbe77a0b4f) in order to be 100% the same as the original library, so I did an amend. Could you please re-review it? 😅

Just to notice the change. Before that change you were able to import the config file like this:

```ts
import {Config} from 'react-native-config';

Config.API_URL; // No TS error, but will crash on compile time
```

And TypeScript won't complain about it, but in runtime it would crash because the actual library doesn't export the object like that.

So with the change now the one and only way to import the config object is:

```ts
import Config from 'react-native-config';

Config.API_URL; // OK
```

Just as we do with vanilla JS with this library.

marcos09 (approved):

Stukz (approved):

PR #212, opened 2022-12-22
Feat: pressable button

tamarasuarezrod:
### Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Documentation content changes

---

### Description

- Create a custom pressable component with a highlight gesture when pressed, this component could be used for buttons, links, icons and any other pressable element. I also added a minimum size of 44px for accessibility purposes.
- Create a simple button wrapped with the previous pressable component.

---

#### Jira board reference

- [Button component](https://www.notion.so/rootstrap/274322b64619490488011b574816bfce?v=bc9fa864b81941d2afd477ff9a6d1680&p=50e1823b37b040edbd3ac12ca374d1c7&pm=s)

---

#### Notes:

- Pressable was released with React Native v0.63, it's explicitly suggested in the [official blog](https://reactnative.dev/blog/2020/07/06/version-0.63#pressable) to start using this component: 
_We expect that most people will build and share components utilizing Pressable under the hood instead of relying on the default experience of something like TouchableOpacity._
You can read [this article](https://blog.logrocket.com/react-native-touchable-vs-pressable-components/#touchableopacity-component) for more information.
---

#### Tasks:

- [x] Tested on IOS
- [x] Tested on Android
- [x] Tested on a small device
- [ ] Tested on a real device
- [ ] Tested accessibility
- [ ] Added tests

---

#### Preview
Screenshot 2022-12-22 at 15 56 03

Screenshot 2022-12-22 at 15 55 52

### DIFF
--- a/src/common/Button/index.tsx
+++ b/src/common/Button/index.tsx
@@ -1,0 +1,18 @@
+import Pressable from 'common/Pressable';
+import React from 'react';
+import { Text } from 'react-native';
+
+import styles from './styles';
+import { ButtonProps } from './types';
+
+const Button: React.FunctionComponent<ButtonProps> = ({ title, disabled = false, ...props }) => (
+  <Pressable
+    containerStyle={styles.container}
+    disabledStyle={styles.disabledContainer}
+    disabled={disabled}
+    {...props}>
+    <Text style={[styles.title, disabled && styles.disabledTitle]}>{title}</Text>
+  </Pressable>
+);
+
+export default Button;
--- a/src/common/Pressable/index.tsx
+++ b/src/common/Pressable/index.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import { Pressable as RNPressable } from 'react-native';
+
+import styles from './styles';
+import { ButtonProps } from './types';
+
+const Pressable: React.FunctionComponent<ButtonProps> = ({
+  containerStyle,
+  disabledStyle,
+  disabled = false,
+  ...props
+}) => (
+  <RNPressable
+    accessibilityRole="button"
+    style={({ pressed }) => [
+      pressed ? styles.pressed : styles.notPressed,
+      styles.commonContainer,
+      containerStyle,
+      disabled && disabledStyle,
+    ]}
+    disabled={disabled}
+    {...props}
+  />
+);
+
+export default Pressable;
--- a/src/common/Pressable/styles.ts
+++ b/src/common/Pressable/styles.ts
@@ -1,0 +1,18 @@
+import { StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  commonContainer: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    minHeight: 44,
+    minWidth: 44,
+  },
+  pressed: {
+    opacity: 0.2,
+  },
+  notPressed: {
+    opacity: 1,
+  },
+});
+
+export default styles;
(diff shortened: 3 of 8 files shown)


src/common/Button/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import styles from './styles';
+import { ButtonProps } from './types';
+
+const Button: React.FunctionComponent<ButtonProps> = ({ title, disabled = false, ...props }) => (
+  <Pressable
+    containerStyle={styles.container}
+    disabledStyle={styles.disabledContainer}
+    disabled={disabled}
+    {...props}>
+    <Text style={[styles.title, disabled ? styles.disabledTitle : {}]}>{title}</Text>
```
ManuViola77, line 14:
```suggestion
    <Text style={[styles.title, disabled && styles.disabledTitle]}>{title}</Text>
```
I think you can do this if you prefer

src/common/Pressable/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children,
+  ...props
+}) => (
+  <RNPressable
+    accessibilityRole="button"
+    style={({ pressed }) => [
+      pressed ? styles.pressed : styles.notPressed,
+      styles.commonContainer,
+      containerStyle,
+      disabled ? disabledStyle : {},
```
ManuViola77, line 20:
```suggestion
      disabled && disabledStyle,
```

same here

ManuViola77 (approved):
Awesome 😍

juanchoperezj (dismissed):
💯

src/common/Pressable/styles.ts
```diff
@@ -0,0 +1,18 @@
+import { StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  commonContainer: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    minHeight: 44,
+    minWidth: 44,
```
asdolo, line 8: Is it needed to force a `minHeight` and `minWidth` for this generic `Pressable` component? I'm thinking of it as a wrapper for `react-native`'s `Pressable` component with an opacity feedback, so maybe having a pre-defined style is not desired.

Same question applies to `alignItems` and `justifyContent`.

src/common/Pressable/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+    ]}
+    disabled={disabled}
+    {...props}>
+    {children}
```
asdolo, line 24: I think it's not needed to explicitly render the `children` here. It'd be included in `props` if we remove the [`destructuring` of it at line 11](https://github.com/rootstrap/react-native-base/pull/212/files#diff-5b89338e4381280d278ab9f6f5013ea37deef797f755684209268a99413703f6R11).

asdolo (commented):
Loving it!

asdolo (approved):

src/common/Pressable/styles.ts
tamarasuarezrod: These props are to follow the suggested accessibility statement which says all pressable buttons should have at least 44px, (you can read more [here](https://www.accessguide.io/guide/large-target-size#:~:text=Large%20buttons%20(44%20px)%20are,interact%20with%20websites%20and%20apps.)). You can always change it if you don't need it but I think it'd be a good guide to have it as a default value in our base.

asdolo (approved):

ManuViola77 (approved):

juanchoperezj (approved):

PR #220, opened 2023-02-02
Create componentes folder template

marcos09:
### Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes
- [X] Documentation content changes

---

### Description

Add support for Template folders Extension VSCode

---

#### Preview

https://user-images.githubusercontent.com/11773865/216385895-915a5d50-3e28-496c-9746-7b3b1ad192f3.mov

### DIFF
--- a/.fttemplates/TS Functional Component/[FTName].tsx
+++ b/.fttemplates/TS Functional Component/[FTName].tsx
@@ -3,10 +3,10 @@ import { Text, View } from 'react-native';
 
 import styles from './[FTName].styles';
 
-const FTName: React.FunctionComponent = () => (
+const <FTName>: React.FunctionComponent = () => (
   <View>
     <Text> REPLACE ME</Text>
   </View>
 );
 
-export default FTName;
+export { <FTName | capitalize> }
--- a/.fttemplates/TS Functional Component/[FTName].tsx
+++ b/.fttemplates/TS Functional Component/[FTName].tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+import { Text, View } from 'react-native';
+
+import styles from './[FTName].styles';
+
+const FTName: React.FunctionComponent = () => (
+  <View>
+    <Text> REPLACE ME</Text>
+  </View>
+);
+
+export default FTName;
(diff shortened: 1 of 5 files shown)


juanchoperezj (approved):

.fttemplates/TS Functional Component/[FTName].tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Text, View } from 'react-native';
+
+import styles from './[FTName].styles';
+
+const FTName: React.FunctionComponent = () => (
```
Comment (aguscha333) [inline .fttemplates/TS Functional Component/[FTName].tsx:6]: I've recently seen components typed this way more often
```suggestion
interface <FTName>Props = {}

const FTName= ({}: <FTName>Props) => (
```
Depending on the version of react that react native is using `React.FunctionComponent` [implicitly receives children as props by default and we might not want that for every component.](https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13#:~:text=FunctionComponent%20provides%20an%20implicit%20definition,children%20provided%20to%20your%20component.)

I don't see a benefit to adding `React.FunctionComponent` but I am open to discussing this and come up with the optimal solution

.fttemplates/TS Functional Component/[FTName].tsx
```diff
@@ -0,0 +1,12 @@
...
+  <View>
+    <Text> REPLACE ME</Text>
+  </View>
+);
+
+export default FTName;
```
Comment (aguscha333) [inline .fttemplates/TS Functional Component/[FTName].tsx:12]: We should consider if default exports is sth that we want to keep or discourage. At MasterClass we avoid them at all cost because of [this](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad)

Same comment applies for the `index.tsx` file
There I would just do:
```js
export { <FTName | capitalize> } from './[FTName]'
```

aguscha333 (commented):
Thanks for adding this! In the future I see us extending this to have a test file as well when we get our testing setup going so adding tests has less friction and devs feel more inclined to do so.

.fttemplates/TS Functional Component/index.tsx
```diff
@@ -0,0 +1,3 @@
+import { <FTName | capitalize> } from './[FTName]'
+
+export { <FTName | capitalize> }
```
aguscha333, line 3: like this is enough since it is not being used inside the file
```suggestion
export { <FTName | capitalize> } from './[FTName]'
```

aguscha333 (approved):

Stukz (approved):

PR #245, opened 2023-09-14
Chore: Apply SemVer to both platforms

juanchoperezj:
### Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [x] Documentation content changes

---

### Description

Both platforms' versions by default were set to 1.0.

**Android** version and build number were handled in the `defaultConfig` and not in each flavor.
**iOS** Have schemes allowing us to edit versions and build numbers per each environment.

These changes allow us to change flavors and scheme versions, asking which environment to use and then what number to bump (major, minor, patch, build).

---

#### Jira board reference

- N/A

---

#### Notes:

-

---

#### Tasks:

- [ ] Tested on IOS
- [ ] Tested on Android
- [ ] Tested on a real device
- [ ] Tested accessibility
- [ ] Added tests

---

#### Preview

https://github.com/rootstrap/react-native-base/assets/17632690/7d0bf887-686d-4855-af78-9ecb2168b879

### DIFF
--- a/rnb-cli/src/tools/rnbv.js
+++ b/rnb-cli/src/tools/rnbv.js
@@ -7,16 +7,19 @@ const semver = require("semver")
 
 const configMap = new Map([
   ["androidPath", "./android/app/build.gradle"],
-  ["iosPath", "./ios/ReactNativeBase/Info.plist"],
+  ["iosPath", "./ios/ReactNativeBase.xcodeproj/project.pbxproj"],
 ])
+let environment;
 
 const androidPath = configMap.get("androidPath")
 const iosPath = configMap.get("iosPath")
 
 const ANDROID_REGEX = /versionName "([.|\d]+)"/
 const ANDROID_REGEX_CODE = /versionCode \d+/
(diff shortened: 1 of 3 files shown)


fernandatoledo (commented):
This is great!!! Maybe it would be useful to add some info to the PR description, the same way you did on the slack message to provide some context on the current situation and proposed solution?

Great work 👏 👏

rnb-cli/src/tools/rnbv.js
```diff
@@ -7,16 +7,19 @@ const semver = require("semver")
 
 const configMap = new Map([
   ["androidPath", "./android/app/build.gradle"],
-  ["iosPath", "./ios/ReactNativeBase/Info.plist"],
+  ["iosPath", "./ios/ReactNativeBase.xcodeproj/project.pbxproj"],
```
juanchoperezj, line 10: We only have one Info.plist, changing version and build numbers here will be applied to all schemes, that's why we now use `project.pbxproj` to edit MARKETING_VERSION and CURRENT_PROJECT_VERSION per scheme.

jpirazusta (approved):
Awesome! 🚀

fernandatoledo (approved):
That description looks great now!!! Thank you for this Juancho 😄

PR #255, opened 2024-02-02
Update scripts/validate-env.ts

fernandatoledo:
### Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Documentation content changes

---

### Description

The development environment was using .env file instead of .env.dev

---

#### Jira board reference

- [Jira_Card_Name](Jira_link_goes_here)

---

#### Notes:

-

---

#### Tasks:

- [ ] Tested on IOS
- [ ] Tested on Android
- [ ] Tested on a real device
- [ ] Tested accessibility
- [ ] Added tests

---

#### Preview

Screenshot 2024-02-02 at 12 20 21

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 ## Getting started
 
 1. Install dependencies: `yarn`
-2. Create a `.env.dev` file in the root directory of the project, based on the `.env.defaults` sample file and the extra constants that you may need. This will be your development env.
+2. Create a `.env.dev` file in the root directory of the project, based on the `.env.example` sample file and the extra constants that you may need. This will be your development env.
 
 You can also create `.env.prod`, `.env.staging`, and `.env.qa` to define environment variables for production and staging.
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 ## Getting started
 
 1. Install dependencies: `yarn`
-2. Create a `.env` file in the root directory of the project, based on the `.env.defaults` sample file and the extra constants that you may need. This will be your development env.
+2. Create a `.env.dev` file in the root directory of the project, based on the `.env.defaults` sample file and the extra constants that you may need. This will be your development env.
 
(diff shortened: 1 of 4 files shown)


asdolo (approved):

asdolo:
Great! Why not to use `.env.dev` for the dev environment instead of `.env` to be consistent? I believe `.env` file is reserved for `react-native-config` as a transition file between the actual environment files and the native one in iOS (`ios/tmp.xcconfig`).

jpirazusta (approved):

fernandatoledo:
You're right, I like that better @asdolo !

asdolo:
Looks great!

Remember to update `Readme.md`. In the `Getting started` section we need to change the second step to say `.env.dev` instead of `.env`. Same with the `Managing multiple environments` section.

Then, in order for this change to work in Android we need to change `.env` `.env.dev` as well in the `envConfigFiles` object at `android/app/build.gradle`.

For iOS I think [it currently searchs for a `.env.dev` file already](https://github.com/rootstrap/react-native-base/blob/42ca274aeb41ce909f8d7d62824739a25f4ffee4/ios/ReactNativeBase.xcodeproj/project.pbxproj#L1115), so I think  it works fine.

Also, when iOS app builds, [it creates a temporal `.env` file](https://github.com/rootstrap/react-native-base/blob/42ca274aeb41ce909f8d7d62824739a25f4ffee4/ios/ReactNativeBase.xcodeproj/project.pbxproj#L347), which is a copy of the actual environment file of the environment being used (`.env.dev`, `.env.qa`, `.env.prod`). **That file is not removed after the build is completed**, so we need to make sure we only use `.env.dev` file to store environment variables for the develop environment and not `.env`, because it will be overwritten.

jpirazusta (approved):

asdolo (approved):

Stukz (approved):

juanchoperezj (approved):

PR #265, opened 2024-05-03
Feat/force update

JUNIOR140889:
### Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [x] Documentation content changes

---

### Description

Force update to minimum version with firebase connection

---

#### Jira board reference

- [Force Update](https://www.notion.so/rootstrap/Force-Update-c45c35b689af460cbe2656ed988241e5)

---

#### Notes:

-

---

#### Tasks:

- [x] Tested on IOS
- [x] Tested on Android
- [x] Tested on a real device
- [ ] Tested accessibility
- [ ] Added tests

---

#### Preview
simulator_screenshot_F78E7A9F-3D68-4765-AEA4-7278BA21C888


src/features/force-update-modal/force-update-modal.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { translate } from 'localization/hooks';
+
+import { isUpdated, openStore } from 'network/services/force-update';
+
+import { styles } from './force-update-modal.styles';
+
+interface ForceUpdateProps {}
+
+const ForceUpdateModal = ({}: ForceUpdateProps) => (
+  <Modal animationType="slide" transparent={true} visible={isUpdated()}>
```
juanchoperezj, line 13: Did you test it on an Android device? 
What would happen if an Android user hits the back button?

src/network/services/force-update.ts
```diff
@@ -0,0 +1,46 @@
+import { Linking, Platform } from 'react-native';
```
juanchoperezj, line 1: What about moving this file to the `hooks` folder?
Even though the Firebase SDK should be making HTTP requests to its server I believe that `network/services` folder is meant for this kind of **[services](https://github.com/rootstrap/react-native-base/blob/main/src/network/services/example-services.ts)**

src/features/force-update-modal/force-update-modal.tsx
JUNIOR140889: I test it in android and the modal keeps showing after pressing the back button.

src/network/services/force-update.ts
JUNIOR140889: In which level should it be the ` hooks`  folder, inside ` network`  or inside `src`?

juanchoperezj: Should be in `src/hooks/`, I checked and the base doesn't have that folder, you will have to create it and add your custom hook there.

src/network/services/force-update.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import remoteConfig from '@react-native-firebase/remote-config';
+
+export const isUpdated = (): boolean => {
+  remoteConfig()
+    .setDefaults({
+      minimum_version: '',
+    })
+    .then(() => remoteConfig().fetchAndActivate())
+    .then(fetchedRemotely => {
```
juanchoperezj, line 12: You can use async/await here

src/network/services/force-update.ts
```diff
@@ -0,0 +1,46 @@
...
+export const openStore = () => {
+  if (Platform.OS === 'android') {
+    Linking.openURL('');
+  } else {
+    Linking.openURL('');
+  }
+};
+
+function compareVersions(local: string, remote: string): boolean {
+  const localComponents = local.split('.').map(component => parseInt(component));
```
juanchoperezj, line 31: The suggestion below will be more descriptive

```suggestion
  const localVersion = local.split('.').map(versionNumber => parseInt(versionNumber, 10));
```

src/network/services/force-update.ts
```diff
@@ -0,0 +1,46 @@
...
+  const remoteComponents = remote.split('.').map(component => parseInt(component));
```
juanchoperezj, line 32: Same a above comment

src/network/services/force-update.ts
```diff
@@ -0,0 +1,46 @@
...
+        const remoteVersion = remoteConfig().getValue('').asString();
+        const localVersion = DeviceInfo.getVersion();
+        return compareVersions(localVersion, remoteVersion);
+      }
+    });
+  return true;
+};
+
+export const openStore = () => {
+  if (Platform.OS === 'android') {
```
juanchoperezj, line 23: Here you can use `Platform.select` to only call `Linking.openURL` 

[Platform select](https://reactnative.dev/docs/platform#select)

src/network/services/force-update.ts
JUNIOR140889: Done

src/network/services/force-update.ts
JUNIOR140889: Done

src/network/services/force-update.ts
JUNIOR140889: Done

src/network/services/force-update.ts
JUNIOR140889: Done

src/features/force-update-modal/force-update-modal.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  useEffect(() => {
+    const checkAndUpdate = async () => {
+      const updated = await isUpdated();
+      if (!updated) {
+        setShowModal(true);
+      }
+    };
+    checkAndUpdate();
+  }, []);
```
juanchoperezj, line 23: You can save this `useEffect` if you move it within the custom hook, I'll get back to you offline about this.

src/features/force-update-modal/force-update-modal.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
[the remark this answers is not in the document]
JUNIOR140889: Done

src/features/force-update-modal/force-update-modal.tsx
JUNIOR140889: Done

src/hooks/force-update.ts
```diff
@@ -0,0 +1,61 @@
...
+    const localVersion = DeviceInfo.getVersion();
+    return compareVersions(localVersion, remoteVersion);
+  }
+  return true;
+};
+
+const openStore = () => {
+  Platform.select({
+    ios: Linking.openURL(''),
+    android: Linking.openURL(''),
```
juanchoperezj, line 41: You can add these values to the env file and use them with the `react-native-config` library

```suggestion
    ios: Linking.openURL(APP_STORE_URL),
    android: Linking.openURL(PLAY_STORE_URL),
```

src/hooks/force-update.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const isUpdated = async (): Promise<boolean> => {
+  remoteConfig().setDefaults({
+    minimum_version: '',
+  });
+
+  const fetchedRemotely = await remoteConfig().fetchAndActivate();
+
+  if (fetchedRemotely) {
+    const remoteVersion = remoteConfig().getValue('').asString();
```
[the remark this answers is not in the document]
JUNIOR140889: It wouldn't be more convenient to put this value and the default minimum version on the env.ts file instead of creating another enum?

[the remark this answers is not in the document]
juanchoperezj: These are different things, we want the enum to be a set of constant values and use it to `getValue`s from the remote config to avoid misspellings or typos.

As for the default `minimun_version` could be in the env file but I don't see the benefit if the default value could be null or a falsy to avoid the function comparison running.

src/hooks/force-update.ts
```diff
@@ -0,0 +1,61 @@
+import { useEffect, useState } from 'react';
+import { Linking, Platform } from 'react-native';
+import DeviceInfo from 'react-native-device-info';
+
+import remoteConfig from '@react-native-firebase/remote-config';
+
+export const useValidateAppVersion = () => {
+  const [showModal, setShowModal] = useState(false);
```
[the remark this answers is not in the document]
JUNIOR140889: Done

src/hooks/force-update.ts
```diff
@@ -0,0 +1,61 @@
+import { useEffect, useState } from 'react';
```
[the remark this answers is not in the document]
JUNIOR140889: Done

src/hooks/force-update.ts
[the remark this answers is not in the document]
JUNIOR140889: Done

src/hooks/force-update.ts
JUNIOR140889: Done

src/hooks/useValidateAppVersion.ts
```diff
@@ -0,0 +1,64 @@
...
+      const updateAvailable = compareVersions(localVersion, remoteVersion);
+      setIsUpdateAvailable(updateAvailable);
+    } catch (err) {
+      console.log(err);
+    }
+  };
+
+  function compareVersions(local: string, remote: string) {
+    if (!local || !remote) {
+      return true;
```
juanchoperezj, line 41:
```suggestion
      return false;
```

src/hooks/useValidateAppVersion.ts
```diff
@@ -0,0 +1,64 @@
...
+        return true;
+      } else if (localVersionComponent > remoteVersionComponent) {
+        return false;
+      }
+    }
+  }
+
+  useEffect(() => {
+    checkUpdate();
+  });
```
juanchoperezj, line 61: Remember to add the dependency array to the useEffect with it will only run once.
```suggestion
  },[]);
```

src/hooks/useValidateAppVersion.ts
```diff
@@ -0,0 +1,64 @@
...
+      const remoteVersion = remoteConfig().getValue(RemoteFields.MINIMUM_VERSION).asString();
+      const localVersion = DeviceInfo.getVersion();
+      const updateAvailable = compareVersions(localVersion, remoteVersion);
+      setIsUpdateAvailable(updateAvailable);
+    } catch (err) {
+      console.log(err);
+    }
+  };
+
+  function compareVersions(local: string, remote: string) {
```
juanchoperezj, line 39: You can convert this function to an arrow function, as you did with the other two functions.
That's to keep consistency with the function definition.

src/hooks/useValidateAppVersion.ts
```diff
@@ -0,0 +1,64 @@
+import { useEffect, useState } from 'react';
+import { Linking, Platform } from 'react-native';
+import Config from 'react-native-config';
+import DeviceInfo from 'react-native-device-info';
+
+import remoteConfig from '@react-native-firebase/remote-config';
+
+enum RemoteFields {
+  MINIMUM_VERSION = '',
```
juanchoperezj, line 9: Enum holds a set of named constants, it is required that you explicitly add here the value tight to the `MINIMUM_VERSION` const.

``` suggestion
  MINIMUM_VERSION = 'minimum_version',
```

If you leave it as an empty string the RemoteConfig library will try to fetch an invalid config.

src/hooks/useValidateAppVersion.ts
```diff
@@ -0,0 +1,64 @@
...
+    const localVersion = local.split('.').map(versionNumber => parseInt(versionNumber, 10));
+    const remoteVersion = remote.split('.').map(versionNumber => parseInt(versionNumber, 10));
+
+    for (let i = 0; i < Math.max(localVersion.length, remoteVersion.length); i++) {
+      const localVersionComponent = localVersion[i] || 0;
+      const remoteVersionComponent = remoteVersion[i] || 0;
+
+      if (localVersionComponent < remoteVersionComponent) {
+        return true;
+      } else if (localVersionComponent > remoteVersionComponent) {
```
juanchoperezj, line 53: One thing here, since the default value of `isUpdateAvailable` is `false` I believe that we don't need this `else if` block.
We will check if the `localVersion < remoteVersion` only, what do you think about this?

juanchoperezj (commented):
Great work! A few minor comments and this should be good to go!

src/hooks/useValidateAppVersion.ts
JUNIOR140889: Done

src/hooks/useValidateAppVersion.ts
JUNIOR140889: Done

src/hooks/useValidateAppVersion.ts
JUNIOR140889: Done

src/hooks/useValidateAppVersion.ts
JUNIOR140889: Done

JUNIOR140889: Done

juanchoperezj (dismissed):
Great work 💯

src/hooks/useValidateAppVersion.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      if (localVersionComponent < remoteVersionComponent) {
+        return true;
+      }
+    }
+  };
+
+  useEffect(() => {
+    checkUpdate();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
JUNIOR140889: I fix the eslint error

src/features/force-update-modal/force-update-modal.styles.ts
```diff
@@ -0,0 +1,34 @@
...
+const styles = StyleSheet.create({
+  centeredView: {
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'center',
+    marginTop: 22,
+  },
+  modalView: {
+    margin: 20,
+    backgroundColor: 'white',
```
fernandatoledo, line 12: Please use colors constants

fernandatoledo (dismissed):

src/features/force-update-modal/force-update-modal.styles.ts
JUNIOR140889: Done

juanchoperezj (approved):

